PR #20, opened 2016-12-31
New content in markdown.md page and lots of cleaning-up of other content

seanmadsen:


docs/api/general.md
```diff
@@ -1,124 +1,78 @@
...
 Interface) that can be used to access and manage data in CiviCRM. The
 API is the recommended way for any CiviCRM extension, CMS module, or
-external program to interact with CiviCRM. This section of the wiki
-gives you an introduction to the basics of the API. This page's children
-dive into the depths of the API.
+external program to interact with CiviCRM.
 
-Why use the API?
-----------------
+Utilizing the API is superior to accessing core functions directly (e.g.
```
ErichBSchulz: Using?

totten:
Overall, I like the improvements to prose, the codeblocks, and symbol typography (backticks).

Regarding the fact that Markdown supports multiple notations for some things (`*` vs `-` for lists; `__` vs `**` for bold; `#` vs `==/--` for headings)... it's kind of funny because it feels like you went the exact the opposite of how I usually do it. :)

My intuition says that we're going to get quickly boxed in if anyone tries to use a richtext/toolbar-enabled editor. For example, I did a quick Google to search find some of Markdown editors with toolbars ("B", "I", etc) -- e.g. Github UI, https://simplemde.com/,  and https://stackedit.io/editor . Comparing them:

   * **Bold**: All three use `**` instead of `__`
   * **Bullets**: Github and StackEdit use `-`. SimpleMDE uses `*`.
   * **Headings**: All three use `#`s

My gut says to always follow the rule of "imitate your surroundings; and if there's nothing nearby, do whatever your tool does." But maybe you have more rationale in this area?

seanmadsen:
Yeah @totten I would definitely like to see more discussion about markdown standards. I raised the issue earlier today in the documentation channel on Mattermost. As well, I wrote a small blurb about it in the wiki page to manage the [content migration](https://wiki.civicrm.org/confluence/display/CRMDOC/Content+migration+from+wiki+to+Developer+Guide). 

As far as the edits-in-question are concerned, I was basically just trying to document what I observed to be a very consistently used standard for markdown syntax in the User Guide, thinking that it was an acceptable basis for precedent, going forward.

Personally the most frustrating aspect of the syntax I observe in the User Guide is the `=`/`-` for level 1 and 2 headings. I would much prefer the `#` syntax since it's far easier to change en masse with regex. It sounds like adopting the hash syntax would be a boon for compatibility with md editors, so I would support this change twofold! But, as nitpicky as it feels, I do want consistent code style between the Dev Guide and the User Guide, which would mean that the adoption of hash syntax would warrant some clean up of the User Guide. I think it's worth it though. 

The second most frustrating thing to me in the User Guide syntax is the sequential numbering. I'd love if we could abandon this. 

All in all, I think this markdown code style issue is a conversation much broader than this specific PR. 

Would you feel comfortable merging as-is, on the basis that the documentation covers existing standards (albeit tacit ones)? Or would you rather I remove the bit about code style?

totten:
Yeah, I'm inclined to merge as-is since (a) it has lots of good changes, (b) it's fairly big and liable to grow stale, and (c) I don't actually have a strong opinion for/against particular formatting.

Not sure if @michaelmcandrew and @joannechester intentionally limited the range of syntax in the User Guide ... it's quite possible that the consistency is just an incidental side-effect of using certain tools to migrate content into Markdown.

nganivet:
- the consistency is definitely because of the migration tools, and then people copying/pasting what was already there.
- in terms of code style I would either follow GitHub's editor syntax (since the guide is on GitHub, will simplify edits by non-techie users) or not suggest any standard.

joannechester:
I much prefer # for headings, but that was not how the migration tool translated the formatting in the user guide.  It does (mostly) use * for italics** for bold and - for bullets, 

I would really like to to use #s for all headings.  IMO it is quicker and extends consistently to heading 3 (and  heading 4 although we try to avoid that in the user guides if at all possible). I have been converting the headings I encounter when I am editing sections of the user guide, but there are a lot of sections.  I am not sure I am up to changing all =/- to #s using a script. 

As for "The second most frustrating thing to me in the User Guide syntax is the sequential numbering. I'd love if we could abandon this."  Are you talking about the numbering used for ordered lists such as a series of instructions listed in order of execution?

If so, then ordered lists had to be numbered in the old system so they came across that way during the migration.  If you are wanting to change to just using "1.  " for all items in a ordered list I can see the advantage in that, however I think non-techies will struggle with it.  I actually hate ordered lists as the placing of images and hyperlinks etc can mess up the numbering quite easily. @seanmadsen ,if you were thinking of a different approach altogether than I would be keen to hear of it.

BTW at bottom of https://wiki.civicrm.org/confluence/display/CRM/Book+style+guide ( which I don't think people can find anymore) you can see the beginnings of my attempt to convey the preferred syntax in a more visual way as a some people won't read lists of what to use.

michaelmcandrew:
+1 for #/## instead of ===/---

Re: bulleted lists. I agree that they are overused in the user guide but you should have seen it before!

There were some chapters which consist nearly completely of bulleted lists (ordered or unordered) when really they should have been paragraphs.

having said that i do think there is a time and a place for sequential numbering - as long as it isn't overused. @seanmadsen maybe you can tell us why you don't like it.

PR #22, opened 2017-01-04
use hash syntax for headings, migrate "developer community" wiki page

seancolsen:


seanmadsen:
This is the first official test of the new workflow for migrating wiki pages, which includes the redirect. Let's see if it works!

docs/documentation.md
```diff
@@ -1,56 +1,65 @@
...
 
-For minor changes you can simply edit the markdown online using GitHub. However, for a better editing experience we highly recommend installing `mkdocs`:
+For minor changes you can simply edit the [markdown] online using GitHub.
+However, for a better editing experience we highly recommend installing
+`mkdocs` as follows.
 
 1.  Obtain the source files for the guide you want to edit
-    1. Find the repository on GitHub (e.g. here is [the repo for this guide](https://github.com/civicrm/civicrm-dev-docs))
-    1. Fork and clone locally. (For more help with Git and GitHub see [Git](git))
+    1.  Find the repository on GitHub (see bottom left of screen)
```
totten: Is this in anticipation of a potential move to different github repo?

docs/markdownrules.md
```diff
@@ -82,38 +71,34 @@ This is a second.
 ```
 
 
+## Headings
 
-Headings
---------
+```md
```
totten: :)

totten:
> This is the first official test of the new workflow for migrating wiki pages, which includes the redirect. Let's see if it works!

If the cron job is updating correctly, it should take effect within 10 min of merge.

docs/documentation.md
seanmadsen: @totten my original motivation was to generalize the instructions to apply equally to the Dev Guide and the User Guide. (When I wrote these instructions originally, I hadn't yet figured out how to get the repo link to show up in the Dev Guide at the bottom left of the screen like it does in the User Guide.) But, now that you mention it, this change done indeed have the added benefit of requiring one fewer edit if/when the repo moves locations.

PR #29, opened 2017-01-06
API: code examples broken or with non-standard style

agh1:
A lot of the code examples in the API pages are bad.  Many have poor style, so they're a bad example for beginners.  Worse, a number are missing closing parentheses, so they won't even work.  This fixes those.

### DIFF
--- a/docs/api/params.md
+++ b/docs/api/params.md
@@ -25,15 +25,22 @@ or by ID.
 Example without sequential:
 
 ```php
-$result= civicrm_api('UFMatch','Get', array('version' =>3, 'uf_id' => $user->uid);
-$contactid=$contact['values'][$result['id']]['contact_id'] );
+$result = civicrm_api('UFMatch', 'Get', array(
+  'version' => 3,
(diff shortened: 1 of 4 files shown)


docs/api/params.md
```diff
@@ -25,15 +25,22 @@ or by ID.
 Example without sequential:
 
 ```php
-$result= civicrm_api('UFMatch','Get', array('version' =>3, 'uf_id' => $user->uid);
-$contactid=$contact['values'][$result['id']]['contact_id'] );
+$result = civicrm_api('UFMatch', 'Get', array(
+  'version' => 3,
+  'uf_id' => $user->uid,
+));
+$contactid = $contact['values'][$result['id']]['contact_id'];
```
ErichBSchulz: this still doesn't look right?

totten: There is a problem in the last line - `$contact` and `$result` are mixed up in the last line. (Pre-existing problem, of course.)

totten:
Sharp eye, @agh1!

PR #32, opened 2017-01-06
indent comments in yml

ErichBSchulz:
background here:

https://chat.civicrm.org/civicrm/pl/9w3jtqc7wprcdfozcaj8r7ag8c


michaelmcandrew:
thanks!

PR #37, opened 2017-01-08
Admonitions and other improvements to markdown documentation

### LINKED ISSUE #36: easier templates for "Note:", "Danger:" info boxes
@ErichBSchulz made some edits recently where he created breakout boxes with icons for "information" and "danger" messages. Here's an [example](https://github.com/ErichBSchulz/civicrm-dev-docs/blob/536dd5b41c42f69080c25a510a719ceeffb732b9/docs/extensions/civix.md) (there are a few in this page). 

I thought this was cool and it got me thinking about how we could make this sort of content more intuitive to create and more visually appealing when rendered. 

I found an extension called [Admonition ](https://pythonhosted.org/Markdown/extensions/admonition.html) that does exactly that. I'd like to play with it a bit to evaluate if it might be worth using in the Dev Guide and/or other guides.

seanmadsen:
Most notably, this PR adds (and documents) support for [admonitions](https://pythonhosted.org/Markdown/extensions/admonition.html), you know... those little break-out boxes like "Note", "Warning", etc. and this change should close #36 

Other smaller tidy stuff included as well, but nothing too major.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,7 @@
-# CiviCRM developer documentation
+# CiviCRM Developer Guide
 
-```bash
-git clone https://github.com/civicrm/civicrm-dev-docs
-cd civicrm-dev-docs
-mkdocs serve
-```
+-   [Read published version](http://docs.civicrm.org/dev/en/master)
+-   [Learn how to edit](https://docs.civicrm.org/dev/en/master/documentation/#how-to-edit)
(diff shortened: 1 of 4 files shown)


michaelmcandrew:
@seanmadsen - I did not know about this feature - very cool!!!

PR #44, opened 2017-01-11
Wiki import "Before you start" and "Recommendations" pages

seanmadsen:
Importing (and cleaning) the following wiki pages: 

https://wiki.civicrm.org/confluence/display/CRMDOC/Before+you+start
https://wiki.civicrm.org/confluence/display/CRMDOC/Recommendations

...both into: `/basics/planning.md`

### DIFF
--- a/docs/basics/planning.md
+++ b/docs/basics/planning.md
@@ -1,0 +1,59 @@
+# Planning your project
+
+Before you start on any code to extend CiviCRM, it is really important
+to discuss your ideas with the community. Here are a few of the reasons
+why this is a good idea:
+
+-   It may have been done already
+-   You'll get suggestions and advice on suitable ways to approach the
+    problem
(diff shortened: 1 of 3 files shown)


seanmadsen:
Re *"you'll"*: linguistically speaking, I'd say this is a contraction, which I think is okay. There was nothing in the old [style guide](https://wiki.civicrm.org/confluence/display/CRM/Book+style+guide) about contractions, and I didn't add anything to my [proposed style guide](https://github.com/civicrm/civicrm-dev-docs/pull/26) about contractions. I think that if we ban contractions, lots of sentences will get pretty weird because words like "it's", "don't" will need to be expanded. And creating a sufficiently detailed style guide to specifically allow and disallow certain contractions seems a bit over the top to me. I think both **"you'll"** or **"you will"** are totally fine.

PR #45, opened 2017-01-11
new page: Prerequisite skills

seanmadsen:
This page will help newbies get up to speed

### DIFF
--- a/docs/basics/skills.md
+++ b/docs/basics/skills.md
@@ -3,23 +3,30 @@
 ## Learn how to *use* CiviCRM
 
 Before diving into CiviCRM development, it's worth mentioning that a solid
-understanding of CiviCRM from the *user's* perspective is very important.
-To that end, please see the
+understanding of CiviCRM from the *user's* perspective can really help. You
+pick up the standard user interface patterns that you can re-use in your work,
+and you might even realise that the UI already provides the functionality you
+wish to develop. To that end, please see the
 [User Guide](https://docs.civicrm.org/user/en/stable/).
 
 
 ## Learn these developer skills
 
-Becoming a proficient CiviCRM developer requires skills in many different
-technologies, some of which are contained within the CiviCRM project
-(e.g. civix, buildkit) and thus covered in detail within this guide. But
-other technologies such as PHP, and Git (which are not CiviCRM-specific) are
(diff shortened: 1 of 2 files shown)


michaelmcandrew:
Hey Sean,

This looks good and really useful.

I have one comment based on something we adopted after a while in the user guide (though I don't think we ever documented) which I think is worth considering here.

We tried to avoid too many constructions that tell people what they should/have to do. Or that there is one way to approach being a developer. Preferring constructions that suggest ways that people can do things. Since we were concerned to avoid things getting proscriptive or feeling intimidating from a user perspective, and we like having a guide that can be consumed by people in different ways.

This typically happened in the user guide because someone would be reading another persons documentation, and think, "actually, I don't do it that way, I do it this way". In these cases, it's much easier to read things like "is it helpful to do X" or "you can do X" rather that "you must to X" (extreme example).

For example "Before diving into CiviCRM development, it's worth mentioning that a solid understanding of CiviCRM from the user's perspective is very important" could be re-written as "Before diving into CiviCRM development, it's worth mentioning that a solid understanding of CiviCRM from the user's perspective can really help. You pick up the standard user interface patterns that you can re-use in your work, and might even realise there is a already way of accomplishing what you are trying to develop in the UI."

Similarly, I like "You don't need to be a pro at everything listed here to get started". But don't like "Becoming a proficient CiviCRM developer requires skills in many different technologies" as much since although it is true to some extent, it can begin to sound like a pretty big hurdle for some people. There are lots of great UI contributions that you could make without knowing any PHP, or performance optimisations that are purely SQL based, for example.

So you could remove the requires sentence and start that section with something like "Below we've outlined the major technologies that CiviCRM is built with. You don't need to be a pro at everything listed here to get started, but it's useful to understand at least of couple of them well, and have a basic understand of how they fit together to create CiviCRM." since this recognises different approaches.

There's definitley a time and place for telling people "the right way" to do something, or how they can avoid traps, but then it is good to explain why.

Let me know what you think,

seanmadsen:
Wonderful feedback @michaelmcandrew. I've made changes to incorporate it. Thanks for the thoughtful approach.

michaelmcandrew:
@seanmadsen - glad it was helpful. thanks for making the changes :)

PR #48, opened 2017-01-11
move front page message about migration up to the top

seanmadsen:


ErichBSchulz:
yes! was thinking same thin

PR #49, opened 2017-01-11
Move/codebase

ErichBSchulz:
I think this is good to go - there is much that needs fleshing out still but this capture all the content from 

https://wiki.civicrm.org/confluence/display/CRMDOC/The+codebase

(love the note/danger blocks)


ErichBSchulz:
cool have done that - hopefully correctly :-)

PR #56, opened 2017-01-14
Migrate wiki page: Debugging

seanmadsen:
An import of the wiki page on debugging

### DIFF
--- a/docs/dev-tools/debugging.md
+++ b/docs/dev-tools/debugging.md
@@ -1,131 +1,53 @@
-# Debugging for developers
-
-!!! warning
-    This is a work in progress.  Orange text needs updating.
+# Debugging
 
 When your code isn't doing what you want it to do, it's time to debug.
 There are lots of options for debugging and there is lots you can do
 without setting up a sophisticated debugging environment.  This chapter
 contains some simple debugging tips and tricks to get you started and
 also instructions on setting up XDebug, which is the recommended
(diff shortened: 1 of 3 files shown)


docs/dev-tools/debugging.md
```diff
@@ -0,0 +1,247 @@
...
+
+```bash
+sudo yum install php-pecl* php-devel php-pear
+sudo pecl install Xdebug
+```
+
+Mac OS X
+
+```bash
+sudo port install php5-xdebug
```
totten: OS X users have many different (but never-the-less common) ways of running PHP (including but not limited to Apple's built-in PHP,  `brew`, MAMP, XAMPP, AMPP, Vagrant, Bitnami, and MacPorts)... but this instruction only applies to MacPorts. Don't think I've encountered OS X users doing Civi on MacPorts for a while.

It kinda sucks, but the best advice I've come up with for OS X users is to identify your specific runtime and search Google, eg `mamp xdebug` or `macports xdebug`.

seanmadsen: Thanks for pointing this out @totten. I added some of what your wrote in the commit below.

seanmadsen:
@totten do you think we can merge this now, given then changes I made to address your feedback?

totten:
👍  Yeah, let's do it. 

I may propose a few more edits.

PR #66, opened 2017-01-22
custom header IDs

seanmadsen:
This PR enables a markdown extension called [attr_list](https://pythonhosted.org/Markdown/extensions/attr_list.html) which gives us the ability to override the default auto-generated ID value on each header. Insert `{:#foo}` after the header text to set an id of `foo`. Then use `http://example.com/mypage#foo` to link to the section of the page beginning with the header. 

Note: I'm not saying we need to start creating custom IDs all over the place. In most cases the auto-generated ones are fine. But the benefit of custom ones is that, for sections we're linking to a lot, we can preserve the links even after edits to the header text.

### DIFF
--- a/docs/markdownrules.md
+++ b/docs/markdownrules.md
@@ -105,10 +105,37 @@ Heading 2
 ---------
 ```
 
-***Convention:*** Each page should have one and only one *Heading 1*,
-which should be at the top of the page as a page title. Other headings within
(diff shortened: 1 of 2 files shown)


seanmadsen:
Ok I'm merging this now since it's a pretty small change and hasn't gotten any comments yet.

PR #72, opened 2017-01-26
Feature/expand architecture

ErichBSchulz:
Most (but not all) of this new content is from the wiki

this page still needs some love (see the comments) but hopefully a step forward!

### DIFF
--- a/docs/core/architecture.md
+++ b/docs/core/architecture.md
@@ -107,11 +107,8 @@ activity is created, or create activities and populating custom fields when a
 pledge is created.
 
 !!! note
-    BAOs typically have both `add()` and `create()` methods. The `add()` method
-    just creates the database record with the DAO `save()` method whereas 'create()' calls
-    subsidiary functions such as creating activities & populating custom fields.
-    The APIs call 'Create' which then calls 'Add' as well as other associated
-    functions.
+    Historically some BAOs had both `add()` and `create()` methods. Current practice 
+    is to favour a single `create()` method.
 
 ### Form
 In general each form page in CiviCRM maps to a file in one of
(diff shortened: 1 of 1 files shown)


docs/core/architecture.md
```diff
@@ -38,33 +40,79 @@ Namespaces are designated with "\".
...
-BAO stands for business access object.  BAOs map to DAOs and extend them with
+BAO stands for business access object
+([example](https://github.com/civicrm/civicrm-core/blob/master/CRM/Event/BAO/Event.php)).
+BAOs map to DAOs and extend them with
 the business logic of CiviCRM.  A lot of the meat of CiviCRM is found in the
 BAOs, for example they have the code that creates follow up activities when an
 activity is created, or create activities and populating custom fields when a
 pledge is created.
 
+!!! note
```
eileenmcnaughton: This is incorrect - they are really really inconsistent - I would probably argue we should move towards removing all the add functions & only having create

seanmadsen:
Ok I'm going to go ahead and merge this now, even if it's a little premature, because we'd like to get a link for this page published so that we can reference it in the wiki (due to this page being an uncommon case of a wiki page with content to migrate but also *sub-pages*). (I can explain more if needed.) More edits can happen in the page once it's published.

seanmadsen:
Thanks for the feedback @eileenmcnaughton !

totten:
I guess this is a bit of tangent, but it'd pretty tempting to move the "API" section above the bao/dao/form stuff. Given all the unit-tests/metadata/bindings (php/cli/ajax), it's usually better for a developer to tap in at that layer -- and drill-down to the others necessary.

docs/core/architecture.md
ErichBSchulz: Thanks @eileenmcnaughton!

PR #77, opened 2017-01-28
New page tree

### LINKED ISSUE #75: Put new page tree into place
We need to re-organize the pages in the Dev Guide a bit. @totten and I have been talking about this and yesterday he gave me the go-ahead to put his [proposed page tree](https://docs.google.com/spreadsheets/d/1ZPq34FuYLBhEzXM2iEGYVZUFLbVZ2CpA3BknLpeHqCQ/edit#gid=1325853905) into place. (Note: I've made some slight adjustments to it, mostly based on other suggestions from Tim in comments.) 

I'm working on this now, and will be moving a bunch of pages around. Although [we don't yet have a good system for page redirects](https://github.com/civicrm/civicrm-docs/issues/20) I'll be forging ahead with some page re-organizing and using inline HTML redirects as a temporary band-aid solution. I'll *also* create a `/redirects/internal.txt` file to keep track of all the page moves so that we we *do* get a proper redirect system, it will be easy to use for these page moves I'll be doing soon. 

This is just a *first push* to get the page tree cleaned up a bit, in light of all the content that will be coming in from the wiki. I'm sure we'll want to adjust the page tree periodically as content evolves, so I'd like to discourage others from making suggestions within in this issue about page organization, unless you feel strongly about it.

seanmadsen:
Ok this should close #75 (at least mostly, and I think the remaining work can be done later and tracked with other issues/PRs)

More details: 

* I opted not to move any *files* around (until we have a good system for redirects). Instead, I just moved the pages around in the nav menu. If I had thought of this option earlier, I would have suggested it. I think it works well for the time being. 
* The nav files has comments that state which pages need to be moved, and the desired new locations for such pages
* In general, lots of comments in the nav file now. I don't see this as good thing to have in there, long there -- these comments are just to help us out in the short term while fleshing out this page tree. 
* Where Tim's proposed page tree recommended pages that do not already exist in the Dev Guide, I added these pages in the nav file with comments. 
* Pages-to-be-created do *not* yet have *files*. (I even removed some placeholder files that @ErichBSchulz created earlier so that this rule is upheld across the board.)  
* I captured all the info from @totten's proposed page tree in google docs. Where Tim had written summaries of the new pages to be created, I added those in comments in the nav file
* I did make some small tweaks to Tim's page tree, but these are easy to change later if we want to discuss them


mkdocs.yml
```diff
@@ -11,52 +11,70 @@ markdown_extensions:
...
-  - Debugging: dev-tools/debugging.md
-- APIv3:
-  - API Intro: api/general.md
-  - API Usage: api/usage.md
-  - API Actions: api/actions.md
-  - API Parameters: api/params.md
+  - Developer Community: basics/community.md    # page-tree = DONE
+  - Prerequisite Skills: basics/skills.md       # page-tree = DONE
+  - Planning Your Project: basics/planning.md   # page-tree = DONE
+- Documentation:
```
ErichBSchulz: i think this meta section could be lower down (even at the end as much as i love docs)

mkdocs.yml
```diff
@@ -11,52 +11,70 @@ markdown_extensions:
...
+  - Writing Documentation: documentation.md     # page-tree = NEED_PAGE_MOVE to /documentation/writing.md
+  - Markdown: markdownrules.md                  # page-tree = NEED_PAGE_MOVE to /documentation/markdown.md
+  - Style Guide: best-practices/documentation-style-guide.md  # page-tree = NEED_PAGE_MOVE to /documentation/style-guide.md
+- Setup:
+  # buildkit: /setup/buildkit.md                # page-tree = NEED_NEW_PAGE  # summary: See Github README for download instructions. Alternatively, follow links and install particular tools as-needed.
+  - Debugging: dev-tools/debugging.md           # page-tree = NEED_PAGE_MOVE to /setup/debugging.md
+- Core Development:
+  - When edit core: core/hacking.md             # page-tree = NEED_PAGE_MOVE to /core/deciding.md
```
ErichBSchulz: how about "When to edit core" - or perhaps even "Introduction" ??

mkdocs.yml
```diff
@@ -11,52 +11,70 @@ markdown_extensions:
...
-- Miscellaneous:
-   - Extension Lifecycle: extend-stages.md
-   - Markdown: markdownrules.md
-   # - hookref-old: hookref-old.md
-- Deprecated:
-   - Building CiviCRM from source: develop-deprecated.md
+  - Using hooks: hook.md                        # page-tree = NEED_PAGE_MOVE to /hooks/usage.md
+- MISC TO REORGANIZE OR DELETE:
+  - Extension advanced patterns: extensions/advanced.md
+  - Extensions files: extensions/files.md
```
ErichBSchulz: I think this could be rolled into `extensions/basics.md `

mkdocs.yml
```diff
@@ -11,52 +11,70 @@ markdown_extensions:
...
-  - Writing Documentation: documentation.md
-- Development Environment:
-  - Debugging: dev-tools/debugging.md
-- APIv3:
-  - API Intro: api/general.md
-  - API Usage: api/usage.md
-  - API Actions: api/actions.md
-  - API Parameters: api/params.md
+  - Developer Community: basics/community.md    # page-tree = DONE
+  - Prerequisite Skills: basics/skills.md       # page-tree = DONE
```
ErichBSchulz: "Useful Skills"

mkdocs.yml
```diff
@@ -11,52 +11,70 @@ markdown_extensions:
...
-  - Architecture: core/architecture.md
-- Miscellaneous:
-   - Extension Lifecycle: extend-stages.md
-   - Markdown: markdownrules.md
-   # - hookref-old: hookref-old.md
-- Deprecated:
-   - Building CiviCRM from source: develop-deprecated.md
+  - Using hooks: hook.md                        # page-tree = NEED_PAGE_MOVE to /hooks/usage.md
+- MISC TO REORGANIZE OR DELETE:
+  - Extension advanced patterns: extensions/advanced.md
```
ErichBSchulz: I think for now could just leave this as last page in extension section

mkdocs.yml
seanmadsen: +1 will fix

mkdocs.yml
seanmadsen: +1 for "When to edit core". That was a typo, I think. Will fix

mkdocs.yml
seanmadsen: Ok sounds good. Will fix.

mkdocs.yml
seanmadsen: Agreed, but I'd like to take care of that later, and get this surface-level page reorganization in place first.

mkdocs.yml
seanmadsen: Yeah, I can see the logic here, but I'd kinda like to leave the Documentation up at front because 

1. it helps order skills in the guide from easy to hard, (e.g. contributing to docs is a great way to get started, like what I'm doing right now)
1. helps encourage developers to get in the habit of documenting early
1. doc will be a very relevant thing in the coming year as CiviCRM works to transition more docs to MkDocs
1. I definitely don't want to put it at the *very* end because that's where the hooks are going, and with "option 4" for the hooks nothing should go below the hooks because they take up so much space in the menu. 

I'm open to moving this section down, but I'd rather leave it as-is for now, since it's something I have an opinion about, and then discuss more in a dedicated issue before making the change you request. Is that okay? If others feel strongly I'll bow out. But I'd like to push this forward for now and sort this detail out later.

seanmadsen:
Ok @ErichBSchulz how about now? You requested 5 changes. I made 3 of them. For the remaining 2, there's 1 that I don't want to make, and 1 that I want to deal with later. Sound okay?

seanmadsen:
Thanks for reviewing, BTW! :slightly_smiling_face:

ErichBSchulz:
4 out of 5 is pretty good! go for it!

PR #89, opened 2017-02-09
extend-stages - removing intro paragraph that doesn't seem to add any real content

seanmadsen:
Hey @GinkgoFJG is this change okay with you? I removed some content that I think (probably?) you wrote, because I felt the page is a little easier to read without it. The rest of the page is all very useful stuff but this intro feels a bit like "fluff" which doesn't really add much. I think the page is more [skimmable](http://www.writethedocs.org/guide/writing/docs-principles/#skimmable) without this intro. Open to rebuttal though.

### DIFF
--- a/docs/extend-stages.md
+++ b/docs/extend-stages.md
@@ -12,7 +12,7 @@ non-profit organizations, we can share our new enhancements and extensions
 &mdash; and build a richer whole for the entire ecosystem.
 
 Of course, this sharing arrangement means that many members of the community
-have a split roles. For example:
+have split roles. For example:
 
 -   **Consumers** want to quickly browse the available
     extensions, pick the ones which look best, and install them. They
--- a/docs/extend-stages.md
+++ b/docs/extend-stages.md
@@ -3,7 +3,7 @@
 This document describes the process of publishing extensions within the CiviCRM
 ecosystem.
 
-## Consumer vs. developer roles
+## Background
 
 The CiviCRM ecosystem is built on the belief that non-profit
 organizations can serve themselves best by collaborating in development of their
--- a/docs/extend-stages.md
+++ b/docs/extend-stages.md
@@ -1,6 +1,25 @@
 # Extension Life cycle
 
-This document describes the process of publishing extensions within the CiviCRM ecosystem.
+The CiviCRM ecosystem is built on the belief that non-profit
+organizations can serve themselves best by collaborating in development of their
+data-management applications.  As staff, volunteers, and consultants for
+non-profit organizations, we can share our new enhancements and extensions --
+and build a richer whole for the entire ecosystem.
+
+Of course, this sharing arrangement means that many members of the community
+have a split role, e.g.
+
+-   Sometimes, we are consumers. We want to quickly browse the available
+    extensions, pick the ones which look best, and install them. We
+    expect these to just work -- both now and going forward (with future
+    upgrades).
+-   Sometimes, we are developers.  We enjoy building great functionality, and we
+    want to invite people to use our products, but we need to juggle the
+    publishing tasks (like testing and maintenance releases) with the goals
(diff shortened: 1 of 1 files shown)


GinkgoFJG:
@totten wrote that. My $0.02 is that it putts the rest of the document into some context, or at least it tries to. I think its intent is to help different classes of people understand which definitions are relevant to them and why. It would seem you are suggesting it could use work :-)

BTW, `git blame` is helpful for figuring out who contributed what.

Cheers!

totten:
Yeah, I wrote most (all?) of this page in consultation with the other Extension WG members.

 * The intro wasn't heavily debated. (Read into that whatever you want.)
 * The tensions between our expectations as developers and as consumers came up over-and-over again in face-to-face and virtual meetings. IMHO, all the rigamarole and definitions and checklists would be superfluous if we didn't have that tension.
 * My main objective: I don't want to get to pulled into a conversation with someone who's taken the view that we should rewrite the doc because consumer-needs outweigh producer-needs, or vice-versa. The whole enterprise requires balance.
 * I see how it'd be a bit fluffy to an incoming developer. The intro softpedals and abstracts a bit.
 * Maybe there's a way to introduce the tensions which resonates more?

seanmadsen:
Ok @totten what about now? 

I put the paragraph back in, but added a little bit of clarifying text. 

In case you're curious, here's what my brain did when I originally approached this page: 

1. Go to 1st paragraph, expecting to get a sense of what the page is about
1. Find 1st paragraph to be be entirely trivial
1. Assume the rest of the page is unlikely to offer me information which I don't already know

I actually glanced at this page a number of times over the past month and sort of "wrote it off" a bit, and it wasn't until I started the process of submitting a formal extension review that I actually understood what this page was about. Of course, if I had initially bothered to continue reading past this intro, I would have understood the page better. But alas I didn't, and so my goal with this edit is to help other sets of fresh eyes take in the gist of the page at a glance.

seanmadsen:
whoops accidentally hit "close". Still open. Seeking go-ahead from Tim before merging.

docs/extend-stages.md
```diff
@@ -1,25 +1,31 @@
...
-    upgrades).
--   Sometimes, we are developers.  We enjoy building great functionality, and we
-    want to invite people to use our products, but we need to juggle the
+have a split roles. For example:
+
+-   **Consumers** want to quickly browse the available
+    extensions, pick the ones which look best, and install them. They
+    expect the extensions to just work &mdash; both now and going forward (with
+    future upgrades).
+-   **developers** enjoy building great functionality, and
```
totten: Consistent caps: `Consumers` vs `Developers`

docs/extend-stages.md
```diff
@@ -1,25 +1,31 @@
 # Extension Life cycle
 
+This document describes the process of publishing extensions within the CiviCRM
+ecosystem.
+
+## Consumer vs. developer roles
```
totten: Overall, I like the changes. 

For a shorter title, maybe just "Background"? That's the title which usually means "Skip past this if your skimming for substance".

seanmadsen:
Yep. Agreed.

docs/extend-stages.md
```diff
@@ -1,25 +1,31 @@
...
 Of course, this sharing arrangement means that many members of the community
-have a split role, e.g.
-
--   Sometimes, we are consumers. We want to quickly browse the available
-    extensions, pick the ones which look best, and install them. We
-    expect these to just work -- both now and going forward (with future
-    upgrades).
--   Sometimes, we are developers.  We enjoy building great functionality, and we
-    want to invite people to use our products, but we need to juggle the
+have a split roles. For example:
```
totten: Oops, should be `have a split role` or `have split roles`

totten: Otherwise, 👍 to merge.

PR #91, opened 2017-02-10
Hook fixes for L, M, N, and O

agh1:
Hit the middle of the alphabet on updating the hook docs.  Some things I standardized:

-   declaring the type of the param, followed by the param name, then a hyphen, then the description
-   putting code blocks in <code>```php</code> rather than indents
-   stripping out references to behavior in pre-4.6 versions

I also did minor grammatical fixes and made sure the code was in proper style.


seanmadsen:
Wow so many good changes in here! Thanks @agh1!! :tada:

PR #93, opened 2017-02-10
More hook fixes

agh1:
A list on hook_civicrm_links was still looking bad, and some code style was still wrong.

I also put the menu of hooks in alphabetical order.


seanmadsen:
Looks great. I like the alphabetization :+1:

PR #103, opened 2017-02-17
Added a requirement for community and official ext

twomice:
@nganivet and I worked out this wording, based on a discussion in the Extensions Working Group. The aim is to help eliminate surprises and increase trust.

### DIFF
--- a/docs/extend-stages.md
+++ b/docs/extend-stages.md
@@ -262,6 +262,7 @@ automated distribution:
 | Distribution	| The project is packaged as a CiviCRM Extension, Drupal Module, Backdrop Module, Joomla Extension, or WordPress plugin	| **Required**	| **Required**
 | Distribution	| The project has a stable version (1.0+; not alpha or beta)	| **Required**	| **Required**
 | Distribution	| A demo site is provided	| *Suggested*	| *Suggested*
+| QA	| The project declares, on the in-app extension management screen, the nature of any changes it makes to existing data or functionality.	| **Required**	| **Required**
 | QA	| The project functions in all CMS's (for CiviCRM Extension)	| *Suggested*	| *Suggested*
 | QA	| An automated test suite is included	| *Suggested*	| **Required**
 | QA	| Project is periodically re-validated with newer versions of CiviCRM and compatibility updates are published	| No	| **Required**


seanmadsen:
Hey @totten can you look over this small change and make sure you agree with it, conceptually? This change originated in yesterday's Extensions Working Group meeting with @GinkgoFJG @nganivet @twomice and me.

seanmadsen:
For some context - we felt that the current criteria for contributed extensions in an official review would technically allow extensions pass even when they mess up people's databases. So we wanted to add another criterion for good measure.

GinkgoFJG:
+1. This is shorter than I thought it would be, but I think it captures the sentiment of our discussion.

nganivet:
Well, Allen and I discussed this. The rest of the document is very terse and we opted for keeping in the same style. We agreed to beef-up the statement if there are any questions or repeat violations of this rule.

GinkgoFJG:
It wasn't a criticism but an observation. Terse is not worse!

totten:
So the expectation is that a reviewer would look for 1-2 sentences in the `info.xml` description?

> ...allow extensions pass even when they mess up people's databases...

Aside: OK, there's lots of ways to mess up a database, and don't expect we can find/imagine all of them, but one particular way is mentioned earlier: "The project does not *override* `PHP`, `TPL`, `JS`, or `SQL` from civicrm-core " -- ie the intent is that extensions must not manipulate the schema of core SQL table. But I'm guessing this patch talks about something else/more-general.

seanmadsen:
Hey @totten, my understanding is that the extensions working group is seeking some go-ahead from you on this PR before adding this additional requirement to the extensions official review (since it [looks like](https://github.com/civicrm/civicrm-dev-docs/blame/682ac84ef3241b5ae6a94dd20350671de67bfe4d/docs/extend-stages.md) you pretty much wrote this whole chapter). Here's the requirement we'd like to add: 

> The project declares, on the in-app extension management screen, the nature of any changes it makes to existing data or functionality.

To answer your clarifying question, yes, the expectation is that a reviewer would look for text in the info.xml description.

And to comment on your aside: yes, we're seeking a new requirement that's a bit more general here. I don't think it overlaps too much with the "does not override" requirement. For example, if an extension seeks to clean up capitalization in people's names, that's not overriding anything, but it *is* making changes to existing data. The risk we're seeking to mitigate is a situation where someone installs an extension that makes unexpected changes to CiviCRM, especially in cases where those changes are irrevocable, like data changes. 

What do you think @totten? Okay to add? Concerns?

seanmadsen:
Hey @totten. We discussed this PR briefly again at the Extensions Working Group meeting and decided to go forward with merging it. If you have any concerns we can address them with future edits.

totten:
@seanmadsen, that's cool. I didn't have any criticisms -- just the clarifications. Carry on.

PR #108, opened 2017-03-06
Update hookref-old.md

johanv:
I'm not sure this code fragment will work. If it really does, it should probably be documented.

### DIFF
--- a/docs/hookref-old.md
+++ b/docs/hookref-old.md
@@ -224,6 +224,8 @@ function modulename_civicrm_pre ($op, objectName, $objectId, &$objectRef) {
   if ($objectName != 'Contribution' || ($op != 'edit' && $op != 'create')) {
     return;
   }
+  // I'm not what will happen when $objectId is NULL, like
+  // when a new contribution is created.
   $contribution_id = $objectId;
   require_once 'CRM/Core/BAO/CustomValueTable.php';
   $my_foo = 'blah';


seanmadsen:
@johanv I'm sorry it's taken me a while to respond to this PR. I've been sitting on it because I haven't really been sure what to do with it. 

There are a number of concerns I have with your proposed change: 

* The file you're changing is placed in the "Content to reorganize" chapter at the end of the book, which means it really needs some large-scale clean up. I haven't looked through this whole page, but at a glance I think it's possible that a decent amount of the content on this page will eventually be deleted in the process of this large-scale clean up. 
* In the comment you've added, it seems like you're missing the word "sure" after "I'm". 
* Most importantly though... Your change seems to be inserting a question (vaguely) into the documentation, which I don't really like. I think if you have a question or concern about the docs, it's best to address raise by creating a [new issue](https://github.com/civicrm/civicrm-dev-docs/issues/new) or asking within Mattermost (prob the [dev](https://chat.civicrm.org/civicrm/channels/dev) channel for this one). Personally, I'm not sure of the correct answer, but I can see your concern about the NULL value. Perhaps you'd like to submit a different PR that changes the code in this example to appropriately deal with the NULL value? I think that would be okay to merge.

I'm going to need some more clarification or adjustments or show of support from other people before I'd feel okay merging this change. Sorry, I'm not trying to be nit-picky. Just trying to maintain a baseline level of quality control. I really do want everyone to be contributing to this guide, and I hope you continue to do so!

I welcome more discussion around this change. If I don't hear anything else, then in 7 days I'll close this ticket without merging.

seanmadsen:
In light of the merge of #118 I'm closing this now.

PR #118, opened 2017-03-19
Removing hookref-old.md

mickadoo:
- Remove `hookref-old.md` and move relevant content to other pages
- Add section on dealing with custom fields to civix page (I'm open to relocating if there's somewhere better)
- Updated some links to the wiki / broken links to blog posts
- Removed warning about user_save for Drupal since I couldn't find any information on the `inCiviCRM`  config variable
- Added Wordpress hooks section


seanmadsen:
Thanks for all this great work @mickadoo! I made one small tweak before merging... I removed the menu entry for `hookref-old.md` from `mkdocs.yml`. Without this step MkDocs wouldn't build for me when testing locally because it complained that the file didn't exist. 

Also I just learned something by looking at your changes &mdash; you removed some h1 elements from the top of a couple pages. I thought this would be a problem, but it's actually not. This is kinda cool! (With our previous theme, it would build the page with no h1 element on the finished html if you didn't specify one in the markdown.) Neat to know that we don't actually have to do that anymore!

mickadoo:
Ah, I had the `hookref-old` in the menu just for testing locally - totally forgot to remove it, thanks.

PR #124, opened 2017-03-28
CRM-20332: modify summaryAction hook to allow user-defined action in third column of action list

monishdeb:


monishdeb:
related core PR is merged.

PR #134, opened 2017-04-20
Separate file system doc

agh1:
Prompted by a [Mattermost conversation](https://chat.civicrm.org/civicrm/pl/9xdgawxbrirjjczchmuapuufyo), I pulled the offending weak explanation of the files directory into its own file.  There's too much to explain to do it clearly in the course of explaining another topic, and once someone understands where the appropriate directories are in their CMS, they get it.


seanmadsen:
This is wonderful!

PR #140, opened 2017-04-27
Update extend-stages.md

totten:
Add more diplomacy.

General thrust: Anyone who makes the journey to reading this doc is probably both consumer+provider. And if somehow they're not... then I want them to empathize and imagine being in both roles.


seanmadsen:
Makes sense

PR #161, opened 2017-05-11
CRM-20037: document added feature for searchColumns hook

monishdeb:


eileenmcnaughton:
@monishdeb this is good but can we get an edit on the first para too - it is no longer true

"The BIGGEST drawback with this hook is that you may need to modify the result template to include your fields. The result files are CRM/{Contact,Contribute,Member,Event…}/Form/Selector.tpl. However, if you use the same number of columns, you can overwrite the existing columns with the values that you want displayed. This is a HACK, but avoids template modification."

seanmadsen:
@eileenmcnaughton are you requesting that the text you quoted be removed? I'm just trying to understand what's needed on this ticket so I can help move it along.

eileenmcnaughton:
I'm hoping @monishdeb will update this to include some more text-tweaks

seanmadsen:
@monishdeb Can you help us finish this PR by making the changes that @eileenmcnaughton is suggesting above? Or if you have another idea for how to move this PR forward let us know.

seanmadsen:
I am merging this as-is for the following reasons: 

* It's been open for almost 3 months
* @monishdeb still has not responded to @eileenmcnaughton's request. 
* The original edit from @monishdeb appears (to me) to be unrelated to the first paragraph. 
* Seems "better in than out"

@eileenmcnaughton if you can describe what sort of edit you'd like to see, I can make it. I don't understand enough about this hook to know what you're talking about. If you're suggesting we delete that text you quoted, then I'd be happy to do that. But from your comment I'm not sure what sort of edit you'd like to see.

PR #166, opened 2017-05-15
Update API schedule option description - 4.3 plus

joemcl:
Updated line 627 to "For use with CiviCRM 4.3 and later," per info from @totten


totten:
Sounds good.

PR #172, opened 2017-06-03
CRM-20600 - Add chapters for AngularJS and AssetBuilder

totten:
Depends: https://github.com/civicrm/civicrm-core/pull/10085/


seanmadsen:
@totten I see that https://github.com/civicrm/civicrm-core/pull/10085 has now been merged. I'd like to get this content into the Dev Guide because it looks really useful. Is there more work to do here?

mkdocs.yml
```diff
@@ -57,6 +57,13 @@ pages:
   - Troubleshooting: extensions/troubleshooting.md # page-tree = DONE
   - Advanced patterns: extensions/advanced.md
 - Framework Reference:
```
seanmadsen: I propose changing this section to the following: 

```yaml
  - AngularJS:
    - AngularJS: framework/angular/index.md
    - AngularJS Quick Start: framework/angular/quickstart.md
    - AngularJS File Names: framework/angular/files.md
    - AngularJS Loader: framework/angular/loader.md
    - AngularJS Modifications:  framework/angular/modify.md

```

Reasons: 
* Use `index.md` for [hackable URLs](https://docs.civicrm.org/dev/en/master/documentation/style-guide/#hackable-urls) (This change also requires a file rename) 
* Use more descriptive menu item labels (as described [here](https://docs.civicrm.org/dev/en/master/documentation/style-guide/#parts-chapters-sections), so that the `<title>` element better reflects the page content).

docs/framework/angular.md
```diff
@@ -0,0 +1,101 @@
...
+
+!!! seealso "Read more about CiviCRM API"
+    A good resource for understanding CiviCRM API concepts is the [APIv3:
+    Intro](/api/general.md).
+
+In the remainder of this document, we'll try to avoid in-depth discussion
+about the internals of AngularJS 1.x or APIv3.  You should be able to follow
+the discussion if you have a beginner-level understanding of both.
+
+
```
seanmadsen: I propose removing this "Additional pages" section because it's a duplication of the listing of pages in the menu and thus is prone to falling out of sync if we add/remove/re-order these pages.

mkdocs.yml
totten: Rebased (to address merge-conflits) and updated to use this structure.

Note: I also renamed "Modifications" to "Changesets". This matches terminology that is used in the code, and it feels a bit more precise.

totten:
I've removed the "WIP" label. Most sections should be presentable.

I don't have a strong opinion on whether to merge now or to wait a bit. There is one page (`framework/angular/changeset.md`) which is incomplete and has a prominent disclaimer at the top. On the other hand, I'm not sure if that page should block the others.

seanmadsen:
Exciting! This stuff looks really good. I have a couple follow-up points for which I'll create tickets.

PR #176, opened 2017-06-07
Add script to locally serve MkDocs site through docker

seanmadsen:
This is a first stab at using Docker to make it easier for people to preview the Dev Guide locally. I'd like some feedback from others about the idea of including this script in this repo (and potentially other repos like the User Guide). @mickadoo @PalanteJon @ErichBSchulz (Docker fans I know) do any of you have thoughts on this? The idea here is that if someone has git and Docker installed locally then they can run this new script `./docker-mkdocs-serve` to preview the book locally and then edit (without needing to install pip, mkdocs, etc). 

Concerns/questions: 

* Would this script even be useful to anyone? Maybe not. If they can install git and Docker it's only a small leap to install the other packages. OTOH, I've seen a *lot* of people have challenges getting MkDocs working locally. *Maybe* there would be *some* people who benefit from using this script?
* Is my approach a good one? I'm new to Docker. The goal here is to use Docker to make it easier for people to run `mkdocs serve`, but am I really using Docker the *right way* here? Not sure about that. Note that there's no Dockerfile here. My approach is to use a [custom public image](https://hub.docker.com/r/seanmadsen/civicrm-mkdocs/) hosted on Docker Hub which I built to have MkDocs and other dependencies that our docs books require.

### DIFF
--- a/docker-mkdocs-serve
+++ b/docker-mkdocs-serve
@@ -1,1 +1,0 @@
-docker run --rm -v "$PWD":/docs -p 8000:8000 -w /docs seanmadsen/civicrm-mkdocs serve --dirtyreload -a 0.0.0.0:8000
--- a/docker-mkdocs-serve
+++ b/docker-mkdocs-serve
@@ -1,0 +1,1 @@
+docker run --rm -v "$PWD":/docs -p 8000:8000 -w /docs seanmadsen/civicrm-mkdocs serve --dirtyreload -a 0.0.0.0:8000
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -71,7 +71,7 @@ The most advanced way to work on a book is to use git to download all the markdo
 1. *(optional)* If you have [Docker](https://www.docker.com/) installed, then at this point you can run the following command and then skip to the "view the book" step below.
 
     ```
-    docker run --rm -v "$PWD":/docs -p 8000:8000 -w /docs seanmadsen/civicrm-mkdocs serve --dirtyreload -a 0.0.0.0:8000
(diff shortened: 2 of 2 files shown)


mickadoo:
I'm not sure how to do it, but it would be nice if the public image had a link to the Dockerfile that was used to build it. Otherwise the script is a bit of a black-box.

I think the script is useful for sure. Having just a single step to get everything should definitely help new doc writers along the way.

I can also see the argument for just putting it in the documentation somewhere instead of a script, since it is just one of the option for getting the environment set up. It also might be customized, for example if you want to change the ports that it's listening on.

It would also be good to know how it'll work on Windows (if we're expecting any Windows devs). If we're not catering to everyone then maybe it would be better placed in a setup section as part of the docs.

seanmadsen:
@mickadoo thanks for the thoughts. I made some adjustments accordingly: 

* [x] *"...would be nice if the public image had a link to the Dockerfile that was used to build it*"
    * I created an [automated build](https://hub.docker.com/r/seanmadsen/civicrm-docker-mkdocs/) for the Docker image (and changed the command to use that new image instead). This way, when you view the automated build on hub.docker.com you see a link to the [GitHub repo](https://github.com/seanmadsen/civicrm-docker-mkdocs) used to generate the build. 
* [x] *"...can also see the argument for just putting it in the documentation somewhere instead of a script"*
    * Agreed. Done. 
* [ ] *"...would also be good to know how it'll work on Windows"*
    * No progress on this front yet

seanmadsen:
I'm going ahead and merging this now, but I hope that we'll continue working on improving this setup, especially with the aim of helping Windows users edit the docs more easily.

Issue #181, opened 2017-06-13
hook_civicrm_alterMailParams contains obsolete discussion


kainuk:
The last part on the hook_civicrm_alterMailParams (https://docs.civicrm.org/dev/en/master/hooks/hook_civicrm_alterMailParams/) is a discussion, not documentation.

PR #182, opened 2017-06-13
Issue #181 removed obsolete discussion

kainuk:


seanmadsen:
Ahh, I see what happened here. Brian Shaughnessy @lcdservices commented on [this wiki page](https://wiki.civicrm.org/confluence/display/CRMDOC//hook_civicrm_alterMailParams) last year, then the page got migrated and the comment came through as part of the page content. I will bring this up with Brian on Mattermost in case he wants to raise this point somewhere else.

PR #201, opened 2017-06-20
Remove "additional pages" section from AngularJS intro page

seanmadsen:
Why? Because these pages are already listed in the menu and I think that listing them in two places will be prone to content errors. 

@totten is this okay with you?


totten:
Removing it seems OK by me.

_Background_: I wanted to be able to get some feedback/testing on the docs before they were merged (esp during the sprint in Germany), and that meant sharing links into the Github interface. The Github UI doesn't provide as much navigation, so it needed an alternative. But now that it's merged+published, that's not a big factor.

(Aside: the sprint experiment seemed succesful: Jo managed to build some new UI's without any previous Angular experience. :) Probably more a credit to Jo than anything else, but the docs were there...)

PR #202, opened 2017-06-20
Add info about AngularJS versions

seanmadsen:
This was confusing to me at first. Adding info to clarify and help others who are new to AngularJS.

### DIFF
--- a/docs/framework/angular/index.md
+++ b/docs/framework/angular/index.md
@@ -8,6 +8,26 @@ administrative screens, and extensions increasingly use AngularJS for
 This documentation aims to explain how AngularJS works within a CiviCRM
 context.
 
+
+## AngularJS versions {:#versions}
+
+* 1.x is called "AngularJS" and has documentation at [docs.angularjs.org](https://docs.angularjs.org)
+* 2.x (and onwards) is called "Angular", has [significant changes](https://angular.io/guide/ajs-quick-reference) from 1.x, and has documentation at [angular.io/docs](https://angular.io/docs/)
+* CiviCRM (as of version 4.7.x) use AngularJS 1.x
+* Within AngularJS 1.x there are [some changes between minor versions](https://docs.angularjs.org/guide/migration)
+
+!!! tip
+
+    To determine the specific version of AngularJS used within your site:
+
+    1. Go to the default Angular *base page* for your site at `http://example.org/civicrm/a`
+    1. Open a browser console
+    1. Evaluate `angular.version` within the console
+
+
+
+
+
 ## Two cultures
 
 CiviCRM is an extensible PHP application (similar to Drupal, Joomla, or
--- a/docs/framework/angular/index.md
+++ b/docs/framework/angular/index.md
@@ -11,9 +11,9 @@ context.
 
 ## AngularJS versions {:#versions}
 
-* 1.x is called "AngularJS" and has documentation at [docs.angularjs.org](https://docs.angularjs.org)
-* 2.x (and onwards) is called "Angular", has [significant changes](https://angular.io/guide/ajs-quick-reference) from 1.x, and has documentation at [angular.io/docs](https://angular.io/docs/)
-* CiviCRM (as of version 4.7.x) use AngularJS 1.x
+* CiviCRM use AngularJS 1.x which has documentation at [docs.**angularjs.org**](https://docs.angularjs.org)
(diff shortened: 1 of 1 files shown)


totten:
Quick responses:

 * +1 on clearly indicating Angular 1.x. The differences between 1.x and 2.x are significant.
 * -1 on pointing to the docs for 2.x. Don't see any value in sending the reader down that rabbit hole. (For comparison: it would be equally silly to say "CiviCRM works best with Drupal 7.x. Here's the documentation for Drupal 5.x and 8.x.")
 * -1 on pointing to the minor-release notes for 1.x. (For comparison: Civi uses jQuery 1.x and mentions jQuery in https://wiki.civicrm.org/confluence/display/CRMDOC/Javascript+Reference , but it doesn't link to minor-release notes.)

Counter suggestion: In the main intro, consistently refer to "AngularJS 1.x" so that the whole thing is framed as 1.x. If you feel strongly that people need to be explicitly warned away from 2.x, then maybe use an admonition.

e.g.

```diff
diff --git a/docs/framework/angular/index.md b/docs/framework/angular/index.md
index d20fb05..cfb8938 100644
--- a/docs/framework/angular/index.md
+++ b/docs/framework/angular/index.md
@@ -1,8 +1,8 @@
 # AngularJS: Overview
 
 AngularJS is a client-side framework for development of rich web
-applications.  The core CiviCRM application uses AngularJS for several
-administrative screens, and extensions increasingly use AngularJS for
+applications.  The core CiviCRM application uses AngularJS 1.x for several
+administrative screens, and extensions increasingly use AngularJS 1.x for
 "leaps" that add or replace major parts of the application.
 
 This documentation aims to explain how AngularJS works within a CiviCRM
@@ -77,9 +77,14 @@ It uses concepts like `ng-app`, "module", "directive", "service", "component", a
 "partial".
 
 !!! seealso "Read more about AngularJS 1.x"
-    A good resource for understanding AngularJS concepts is [the
+    A good resource for understanding AngularJS 1.x is [the
     official AngularJS tutorial](https://code.angularjs.org/1.5.11/docs/tutorial).
 
+!!! warning "Caution: AngularJS 1.x and 2.x are different frameworks."
+    CiviCRM specifically uses AngularJS 1.x. AngularJS 2.x is a
+    significantly different framework, and this documentation
+    only speaks about 1.x.
+
 The second interaction comes when the AngularJS application loads or stores
 data.  This uses the CiviCRM API.  Key concepts in CiviCRM API include
 "entity", "action", "params", the "API Explorer", and the bindings for PHP/Javascript/CLI.
```

seanmadsen:
Agreed re: minor release notes. Fixed.

I'd like to push back on your other point a bit. I made some tweaks which I think will "avoid sending people down that rabbit hole", but the reason I mention angular.io is actually for this very reason &mdash; to help people understand that angular.io is a rabbit hole they should not go down! It's easy to end up in that rabbit hole when you're trying to do something like "learn angular". For example, look at the fact that Joe [added a link to the Dev Guide for "AngularJS docs"](https://github.com/civicrm/civicrm-dev-docs/pull/105/files) which incorrectly pointed to angular.io. I reviewed this change and approved it (but have since corrected it.) The fact that both of us thought *"angular.io is the right place to go"* I think demonstrates that we need to provide a little extra clarity here. If a reader ends up at angular.io they'll have very little indication that they're at the *wrong* place. To extend your comparison, [drupal.org's docs](https://www.drupal.org/documentation) begin from the outset with language that gives the reader a clear choice between version 8 and version 7, and both versions are documented on the same site. The confusion between Angular vs AngularJS seems seems important enough to me to actively point readers *away* from angular.io. 

What do you think about this PR now that I've made these adjustments?

seanmadsen:
Also to respond to your suggestion of mentioning 1.x in more places, I don't think we need to because if the reader understands "AngularJS" = "1.x" and "Angular" = "2.x" (and if we stick to that language), then I think we're doing enough to communicate this concept.

seanmadsen:
@totten can you give this another look now that I've adjusted it a bit? I'd like to either merge this, adjust further to suit your wishes, or close.

seanmadsen:
@totten I am merging this now for the following reasons: 

* I *think* I've addressed your underlying concerns with my changes
* It's been 4 weeks and I haven't heard from you
* We can continue to make more changes after this is merged

PR #203, opened 2017-06-23
WIP: Add testing documentation

mickadoo:
This PR migrates all testing information from both the old Wiki and some blog posts that I could find.

It is marked as WIP since @stevekessler is also working on the testing docs, so hopefully this PR will either be a joint one or will provide some help for Steve with his changes.


seanmadsen:
@mickadoo this is great stuff!! I made some small adjustments (shown above).

seanmadsen:
@stevekessler I'm looking forward to merging your work with Michael's. Here are the simplest options I see for how you can proceed from here. 

1. Make a separate PR (in which case *I'll* do the work of merging your changes with Michael's changes.)
1. Tell me to merge this PR soon so that you can pull from master and base your changes off of that. (Choose this one if you think it might be a while before you finish your changes.)

Up to you.

seanmadsen:
Oh @mickadoo one more question: were there any *wiki pages* which you *completely* migrated during these changes? Because if so, we should list them in the redirects file and include that change as part of this PR. If you *partially* migrated a wiki page (and the wiki page still has useful content on it that should get migrated somewhere else in the Dev Guide), then no need to add a redirect.

docs/tools/phpstorm.md
```diff
@@ -27,4 +27,39 @@ That's it. You can now use this code style on all future CiviCRM-related project
 
 TODO
 
+## Running automated tests from within PhpStorm {:#testing}
+
```
mickadoo: this makes sense, there were some changes from the sprint that I hadn't seen so didn't notice this dedicated page before.

mickadoo:
@seanmadsen I tried to pull information from a few sources, looking at my history from yesterday I used some information from these sites:

-https://wiki.civicrm.org/confluence/display/CRMDOC/Testing
-https://wiki.civicrm.org/confluence/display/CRM/Writing+a+PHPUnit+testcase+HOWTO
-https://github.com/civicrm/civicrm-buildkit/blob/master/doc/daily-coding.md#upgrade-tests
-https://github.com/civicrm/org.civicrm.testapalooza
-https://github.com/civicrm/civicrm-core/tree/master/tests
-https://wiki.civicrm.org/confluence/display/CRM/CiviCRM+Unit+Testing+basic+information
-https://wiki.civicrm.org/confluence/display/CRMDOC/Tests+in+phpstorm

Unfortunately there's not a 1-to-1 mapping of each page migration. I wanted to be able to understand what I was migrating, and part of that was separating it into the structure that I've put forward in this PR.

seanmadsen:
@mickadoo thanks for listing all those pages. This is really awesome. 

As much as possible, I'd like to avoid situations where we have duplicated documentation across multiple sources. So instead of *copying* content into the Dev Guide, I'd like to be *moving* it (and adding some sort of link which guides readers from the old place to the new place). As I see it, this approach will (1) keep things cleaner, (2) make content easier to maintain, and (3) make other content easier for readers to discover. For example, when I put buildkit docs into the Dev Guide, I made a PR to [remove them from the buildkit repo](https://github.com/civicrm/civicrm-buildkit/pull/332). I've done this with some markdown files in core too. With the wiki, I think the redirect approach suffices because you can effectively no longer reach the old page (even though technically the content still exists in the wiki). 

I think it's okay if there's not a 1-to-1 mapping. Like, for example, with the the "Testing" wiki page which maps to multiple Dev Guide pages, I'd say we should redirect that to the "Testing Setup" page, and readers can find their way to the other pages from there. The important thing I'm after here is just to tidy things up by removing those duplicate sources. *But* we don't want to remove them yet if there's still some useful content in those other pages. So for each of those links above, did you migrate *all* the useful stuff in those pages? Or just *some*? (Ok to abandon some non-useful content of course.)

mickadoo:
@seanmadsen I had a look at the 4 wiki pages I went through:

https://wiki.civicrm.org/confluence/display/CRM/CiviCRM+Unit+Testing+basic+information

- Migrated useful stuff
- Omitted some code samples

https://wiki.civicrm.org/confluence/display/CRMDOC/Tests+in+phpstorm

- Kept basic steps for setting up tests
- Omitted screenshots (some of which are outdatted)

https://wiki.civicrm.org/confluence/display/CRMDOC/Testing

- Kept most content but split it
- Omitted sample console output
- Move CI table to paragraph inside CI page

https://wiki.civicrm.org/confluence/display/CRM/Writing+a+PHPUnit+testcase+HOWTO

- Kept info on being careful about test data
- Omitted info on convenience test methods
- Omitted stuff about "figuring out a test case" 
- Omitted XML data set samples
- Omitted sample empty test
- Removed sample code for inserting data

I'd say that info on writing a PHPUnit test case could be improved, but I think the best way would be to have a consistent style for the real tests which people will be using as an example.

In general some of the split files are a bit small, but I think this is probably because they need a bit more love, instead of a sign that they should be all lumped together. It would be nice to have a consistent structure for the PHPUnit, Selinium, Javascript and Upgrade test sections, with instructions for how to run them and a section on best practices. Unfortunately that content wasn't there in the original so they're a bit bleak now.

seanmadsen:
@stevekessler said [here](https://chat.civicrm.org/civicrm/pl/i8fpqco3r3gb5d7tgm43kyq6jo) that we can go ahead and merge this and then he'll base his changes off of it. I'm going to give this another look-over and then merge it shortly.

seanmadsen:
@mickadoo I'm redirecting all four of those wiki page to their relevant sections in the Dev Guide. This means that the content you omitted will be lost. I scanned over it and I think that's fine. You got the useful bits. 

@stevekessler For your work, if you want to look over those wiki pages that Michael used in migrating this content, you can use the links below: 

* https://wiki.civicrm.org/confluence/display/CRM//CiviCRM+Unit+Testing+basic+information
* https://wiki.civicrm.org/confluence/display/CRMDOC//Tests+in+phpstorm
* https://wiki.civicrm.org/confluence/display/CRMDOC//Testing
* https://wiki.civicrm.org/confluence/display/CRM//Writing+a+PHPUnit+testcase+HOWTO

Note the redundant forward slashes above. That's how you access a wiki page after it's been redirected to the Dev Guide.

PR #204, opened 2017-06-25
Strict markdown standards for internal hyperlink URLs

seanmadsen:
I'm proposing this change which will add **new markdown code standards** which are very specific about how content editors should write the URL for internal hyperlinks. 

### Why

Only recently have we begun to [re-organize pages](https://github.com/civicrm/civicrm-dev-docs/commit/e2ab63816013208e241bc081785222ea1246f76f). Turns out MkDocs is not as good as I thought about detecting broken links. It only finds them when they're formatted `[like](/this.md)`. So today I did some work to [automatically reformat](https://github.com/civicrm/civicrm-dev-docs/commit/f01992b8cf5c1f581dcf7a852330eb7b594900c7) links, and in the process I fixed a bunch of broken links in: b75a88352a685c4ba2dd3640aff6853cc1cdf216 a3362299d49b1f26f8c7762970008de79eba3747 bd2f7683cf4a65139bfbcc138fd6e1529f52c9b8 c0b79cf4a336995e57dadb82a410526634be83a9 865af29411c2ab1d28623a6a49da2b03c92fb831 (Sorry, I should have put all that into a PR for easier reference).

So to be clear (because of my work today) the Dev Guide already conforms to these markdown standards. The User Guide does not, but I think it'd be pretty easy to bring it into conformance (using some similar regex work).

I'd like some feedback from others before adding a new markdown code standard. Comments welcome from anyone. Maybe @totten @michaelmcandrew @mickadoo @PalanteJon @nganivet will have thoughts.

### DIFF
--- a/docs/documentation/markdown.md
+++ b/docs/documentation/markdown.md
@@ -26,10 +26,41 @@ The remainder of this page will focus on the markdown syntax which can be used w
 
 To maintain some consistency and peace of mind for documentation content editors, we've [agreed](https://github.com/civicrm/civicrm-dev-docs/issues/43) to *recommend* the following syntax as markdown code standards. These are not hard rules though.
 
+### General code standards {:#general-standards}
+
 * **Line length:** write long lines (i.e. one line per paragraph) and set your text editor to view them with a "soft wrap".
 * **Ordered lists:** use `1.` as delimiters.
 * **Unordered lists:** use `*` to delimiters.
 * **Headings:** use hashes like `## Heading 2`.
+
+### Internal URL standards {:#internal-url-standards}
+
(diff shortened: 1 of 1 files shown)


mickadoo:
If mkdocs can detect broken internal links using this method then I'm all for it. I did like having the style of having the link reference at the bottom since it could save the need to update multiple spots if the internal link changed, but I think it's a minor thing and having proper checks if internal links are valid is the better choice.

seanmadsen:
I would have liked to receive more feedback on this, but it's been 10 days and only 1 comment (in support) so I'm going ahead and merging this.

PR #205, opened 2017-06-25
Autogenerate hooks summary (with bash)

mickadoo:
## Overview

Currently we have no overview page for all the hooks like the one that existed on the wiki site. This PR will add a script to generate this overview page. It can then be called as part of the publishing process

## Before

There was no overview page. Creating one would require manual update

## After

A new script has been added that will automatically create a page to show the full list of hooks, with links to each hook description.

## Notes

- I removed some legacy scripts since the hook pages have been migrated.


seanmadsen:
@mickadoo and I chatted and decided that for now we will move forward with #206 instead of this PR, but also we will pursue some other approaches (e.g. python) as well.

PR #206, opened 2017-06-26
Autogenerate hooks summary (with Symfony CLI tool)

seanmadsen:
@mickadoo when I looked at your PR #205 I got excited about this project, but I really wanted to see *more* on that summary page. I wanted to see a short description of each hook, and I wanted the hooks to be categorized. I was thinking that those tasks would be a bit cumbersome in bash so I started playing around with it in PHP. Then, well, I got pretty carried away. Yeah... so I hope you don't mind that I took something very cool that *you* did and then went a different direction with it. Sorry! 

What do you think of this though? Check out the [auto-generated summary page](https://github.com/seanmadsen/civicrm-dev-docs/blob/7a9fd62e0e3c6e55cd932cb1b27c447994309663/docs/hooks/list.md).

Mostly this PR is so giant because I ended up doing small tweaks to almost all the hooks files in order to get to a point where I could extract a nice, clean summary out of each of them. 

I'm interested in seeing what else we can do with auto-generated documentation. I don't have anything too specific in mind at the moment, but I have a gut feeling that we'll probably end up wanting to do more stuff like this in the future... hence the somewhat more sophisticated docs-generating infrastructure with a symfony CLI tool here. It could still use some refining though (e.g. using `__DIR__` all over the place), but I kind of like the idea of laying some solid groundwork here that we can build on top of more easily (than bash) if/when we end up wanting to do more complicated things. 

I'm open to other approaches though. What are your thoughts about this?


docs/hooks/hook_civicrm_alterMenu.md
```diff
@@ -1,25 +1,24 @@
...
 !!! note "Comparison of Related Hooks"
     This is one of three related hooks. The hooks:
 
     -   [hook_civicrm_navigationMenu](/hooks/hook_civicrm_navigationMenu.md) manipulates the navigation bar at the top of every screen
-    -    [hook_civicrm_alterMenu](/hooks/hook_civicrm_alterMenu.md) manipulates the list of HTTP routes (using PHP arrays)
+    -   [hook_civicrm_alterMenu](/hooks/hook_civicrm_alterMenu.md) manipulates the list of HTTP routes (using PHP arrays)
     -   [hook_civicrm_xmlMenu](/hooks/hook_civicrm_xmlMenu.md) manipulates the list of HTTP routes (using XML files)
 
 
+## Availability
```
mickadoo: I like the idea of separating this

mickadoo:
@seanmadsen this is a big improvement in terms of structure of the summary page. It's a good idea to use the mkdocs.yml file to fetch the categories. Plus with the changes you did on the hook files I think they're a lot more orderly now.

One thing we might want to consider is that we're adding another language dependency here. Previously editing the docs didn't require you to have PHP installed at all, and now if you want to do the full process when working on hooks you need to install PHP, composer and run composer install before running the command. For most developers it won't be a problem, but just something I think we should consider.

I started out with my simple script by going for PHP, but then reconsidered and wrote it in bash instead. I'm sure it would be possible to use your new changes with the structure and replace the PHP stuff with bash, but I'd like to hear what you think.

seanmadsen:
@mickadoo Yeah I hear you on the drawback of a new language dependency. 

I can see how adding the summary text wouldn't be too hard in bash, but what about grouping by category. Do you have ideas for how to do that in bash that wouldn't be too ugly (or prone to errors if we re-organize mkdocs.yml)? I really do like having them categorised.

seanmadsen:
More thoughts... I think there are two angles to consider here: 

1. the immediate task at hand (generating a hooks summary page)
1. the long-term usability/maintainability/consistency of *all* our auto-docs tools (of which this tool would potentially just be the first). 

Here is my take on PHP vs bash for each of those angles

1. immediate task: it's hard for me to imagine a circumstance in which someone would want to edit the hooks summary page but not already have PHP + composer installed. In the unlikely even that this happens, then the person who merges the PR could run the script to re-generate it. 
2. long-term: if we start using this PHP approach for other things too we could easily end up in a situation where people *do* want to edit the auto-generated docs but lack the tools to do so (e.g. User Guide). So perhaps PHP is not the right tool for creating auto-docs systems, generally speaking. PHP loses some points here for "usability". But I think bash loses *more* points for maintainability (and also loses some points in "usability" because of the additional complexity of adding CLI help). I also briefly considered python since mkdocs already depends on that, but figured "maintainability" would be a challenge given the lower number of civi devs experienced with it. It might even be worth taking a step *further* back and asking whether there are other tools for building static docs sites which would suit our needs better than mkdocs (e.g. interesting to see how the Docker project is using Jekyll to build their docs). But probably that's a bigger ball of wax than either of us want to get into right now.

mickadoo:
I was messing around looking at how to group by category like [here](https://github.com/mickadoo/civicrm-dev-docs/blob/de5ca423ee043cfaf7908a61d60019fd6320c5bf/bin/hook-make-list.sh) but I can see the point about how it might not be maintainable.

I'm happy to put more time into the bash script if you think it's worth it. Otherwise we can go with the php and symfony approach, I'll leave it up to you.

seanmadsen:
@mickadoo and I chatted and decided that for now we will move forward with this PR (instead of #205), but also we will pursue some other approaches (e.g. python) as well.

PR #208, opened 2017-06-30
Python script for hooks summary

mickadoo:
## Overview

We've gone through a few options for creating a summary script. 

- Bash is short and sweet but may be difficult to run on Windows. As with Python it's less open to extension. 
- The Symfony command is extensible and in a familiar language to most CiviCRM developers, but it adds another language dependency to create the environment to edit the docs. It also add more to the setup, potentially requiring users to install the correct version of PHP, composer and run `composer install`.
- Python is the language mkdocs uses so it makes it simpler for setup to use it in scripting too. When mkdocs releases v1.0.0 with the extensions capabilities it should also be an easier integration process if our scripts are written in Python.

## Before

We used the Symfony console and a custom command to generate the hooks summary page markdown

## After

We use a python command to generate the hooks summary page markdown


seanmadsen:
This is great! :grinning: You win :trophy:

PR #212, opened 2017-07-05
Fixed misnamed hook

GinkgoFJG:
Note this change renames the file as well. A redirect is probably needed to preserve old links, but I'm not sure how this is managed.


seanmadsen:
Thanks for this @GinkgoFJG 

I made some necessary adjustments (including the redirect, as you mentioned) which you can see in my three commits above.

GinkgoFJG:
Thanks for getting into the corners I couldn't reach, @seanmadsen!

PR #217, opened 2017-07-12
Highlight civicrm.settings.d

seanmadsen:
### Before

* Info about `civicrm.settings.d` was a bit buried

### After

* Content in the civibuild page is changed to make it easier for readers who are *skimming* to identify this concept.
    * Heading changed from "civicrm.settings.php"  to "civicrm.settings.d folders" makes it clearer that this is a *different* concept from the simple per-site config available with civicrm.settings.php. 
    * New content copied in from [here](https://github.com/civicrm/civicrm-buildkit/blob/master/app/civicrm.settings.d/README.txt)
    * Content put at the front of the section which makes it easier to get the gist of it by just reading the first sentence. 
* Tip added to the debugging page which points readers to this content

Is this okay with you @totten ?

### DIFF
--- a/docs/tools/civibuild.md
+++ b/docs/tools/civibuild.md
@@ -302,24 +302,6 @@ Civibuild will check the following `civicrm.settings.d` folders.
     
 The `$PRJDIR/app/civicrm.settings.d/` also contains some [example configuration files](https://github.com/civicrm/civicrm-buildkit/tree/master/app/civicrm.settings.d). For more advanced logic, one can look at the global `$civibuild` variable or at any of the standard CiviCRM configuration directives. 
 
-There are a few CiviCRM settings which are commonly configured on a per-server
-or per-workstation basis. For example, civicrm.org's demo server has ~10
-sites running different builds (Drupal/WordPress/CiviHR),
-and visitors should not be allowed to download new extensions on any of those
-sites. However, on the training server, trainees should be allowed to download
-extensions. As discussed in
-[Override CiviCRM Settings](https://wiki.civicrm.org/confluence/display/CRMDOC/Override+CiviCRM+Settings),
-this setting (and many others) can be configured in civicrm.settings.php.
-
-The `civicrm.settings.php` is created automatically as part of the build. One
-could edit the file directly, but that means editing `civicrm.settings.php`
-after every (re)build. The easiest way to customize the settings is to put
-extra `.php` files in `/etc/civicrm.settings.d` &mdash; these files will be loaded
-on every site that runs on this server (regardless of how many sites you
-create or how many times you rebuild them).
-
-
-
 
 ### settings.php; wp-config.php {:#settings-cms}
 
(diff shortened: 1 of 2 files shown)


docs/tools/civibuild.md
```diff
@@ -262,7 +262,29 @@ There are four variations on rebuilding. In order of fastest (least thorough) to
...
+```
+
+Any settings which you would typically put in your site's `civicrm.settings.php` file can go into a php file (you choose the file name) in a `civicrm.settings.d` folder. Further, there are multiple `civicrm.settings.d` folders. Settings files from these directories will be loaded in the following order:
+
+1. `$PRJDIR/app/civicrm.settings.d/`
+1. `$PRJDIR/app/config/$TYPE/civicrm.settings.d/`
+1. `/etc/civicrm.settings.d/`
+1. `$SITE_DIR/civicrm.settings.d/`
+
+If a settings file has the *same name* as one which has already been loaded, it will be skipped.
```
seanmadsen: @totten did I get this right? (load order and skipping files based on file name)

totten: I think that leads to a backwards interpretation. Also, it's easy to confuse order-of-execution (based on filename) and order-of-override (based on folder).

The dynamic might be a bit clearer if we label the folders by their intended use, ie

 1. `$PRJDIR/app/civicrm.settings.d/` (General defaults provided by upstream buildkit for all civibuild sites)
 1. `$PRJDIR/app/config/$TYPE/civicrm.settings.d/`  (General defaults provided by upstream buildkit for specific types of sites)
 1. `/etc/civicrm.settings.d/`  (Overrides provided by the sysadmin for the local server)
 1. `$SITE_DIR/civicrm.settings.d/` (Overrides provided for a specific site/build)

For concrete example, suppose we have these files:
 * `$PRJDIR/app/civicrm.settings.d/200-two.php`
 * `$PRJDIR/app/civicrm.settings.d/300-three.php`
 * `/etc/civicrm.settings.d/100-one.php`
 * `/etc/civicrm.settings.d/300-three.php`

Then we would execute/load in this order:
 * `100-one.php` (specifically `/etc/civicrm.settings.d/100-one.php`; this is the only version of `100-one.php`)
 * `200-two.php` (specifically `$PRJDIR/app/civicrm.settings.d/200-two.php`; this is the only version of `200-two.php`)
 * `300-three.php` (specifically `/etc/civicrm.settings.d/300-three.php`; the system configuration in `/etc` overrides the stock code in `$PRJDIR/app/civicrm.settings.d`)

totten:
Big +1 on moving the explanation from `.txt` to `.md`.

The new heading seems reasonable -- and I appreciate the nuance of it.

If the content is moved into `civicrm-dev-docs`, then the README.txt stubs should point to the docs. You're probably thinking of that already... I just mention because I think we should still have stubs for the people whose path to understanding begins with browsing the source tree.

If you're going to put in some new introductory text, then you might as well kill the introductory text I had in there.

My main concern is about making the overall document too long/overwhelming. IMHO, it's enough to say "You can add *.php files to `/etc/civicrm.settings.d` etal." The finer points of overloading behavior are more advanced concerns.

seanmadsen:
@totten I'm pretty sure I've addressed your feedback with my two latest commits. Merging now. Happy to make further adjustments in subsequent PRs.

PR #221, opened 2017-07-16
CRM-20845: document alterMailingRecipients hook

monishdeb:


docs/hooks/list.md
```diff
@@ -84,6 +84,7 @@ This is an overview list of all available hooks, listed by category.
 * **[hook_civicrm_alterMailContent](/hooks/hook_civicrm_alterMailContent.md)** - called after getting the content of the mail and before tokenizing it.
 * **[hook_civicrm_alterMailer](/hooks/hook_civicrm_alterMailer.md)** - called when CiviCRM prepares an email driver class to handle outbound message delivery.
 * **[hook_civicrm_alterMailParams](/hooks/hook_civicrm_alterMailParams.md)** - called when an email is about to be sent by CiviCRM.
+* **[hook_civicrm_alterMailingRecipients](/hooks/hook_civicrm_alterMailingRecipients.md)** - called twice before and after constructing mail recipients.
```
JoeMurray: could you change 'called twice before and after' to 'called twice, once before and once after' ? Thx

docs/hooks/hook_civicrm_alterMailingRecipients.md
```diff
@@ -0,0 +1,44 @@
+# hook_civicrm_alterMailingRecipients
+
+## Summary
+
+This hook is called to allow the user to alter the mailing recipients after they have been constructed.
+
+### Notes
+
+!!! Note
+    This hook is called two times at the start and end of mail recipients
```
JoeMurray: 'two times, once at the start and once at the end'

aydun:
The function name in the example is wrong: `mymodule_civicrm_alterMailingLabelParams`  should be `mymodule_civicrm_alterMailingRecipients`

docs/hooks/hook_civicrm_alterMailingRecipients.md
```diff
@@ -0,0 +1,42 @@
...
+
+```php
+    function hook_civicrm_alterMailingRecipients(&$mailingObject, &$params, $context);
+```
+
+## Parameters
+
+-   object `$mailingObject` - reference to CRM_Mailing_DAO_Mailing object
+-   array `&$params` - the mailing parameters used to filter mailing recipients
+-   string `$context` - contain 'pre' or 'post' value to indicate when the hook is fired
```
totten: In the `$context`, I like how it provides concrete examples.

I think it's good do that with all array and string parameters -- so, *what keys and values are expected in `$params`*?

JoeMurray: +1 re providing more info on possible values in $params here. @monish when you're back next week can you find an authoritative source for content of $params...likely where they are consumed or perhaps marshalled somewhere in a complete UT?

eileenmcnaughton:
@seamuslee001 seems I can't ....

seamuslee001:
Merging

PR #222, opened 2017-07-17
Add in Coding Standards Docs for PHP and SQL

seamuslee001:
Ping @seanmadsen @xurizaemon Chris are you able to check the SQL one to see if it makes sense i know that you did a talk on writing proper SQL code at the code sprint in Fort Collins last year and just want to make sure its consistent, Sean can you check if i have got the style / phrasing right i think so but just would appreciate a good review.

### DIFF
--- a/docs/standards/sql.md
+++ b/docs/standards/sql.md
@@ -1,13 +1,23 @@
 # SQL Coding Standards
 
-When writing SQL developers should ensure all variables are passed into SQL safely and securely.
+When writing SQL, it is very important that developers protect against [SQL injection](https://en.wikipedia.org/wiki/SQL_injection) by ensuring that all variables are passed into SQL safely and securely.
 
-In CiviCRM this is done using the inbuilt parameterization tools.
+This page describes the inbuilt parameterization tools available for safely executing SQL.
+
+## `CRM_Core_DAO::executeQuery` {:#executeQuery}
 
 ```php
-$name = 'John Smith';
-$optedOut = 0;
-$result = CRM_Core_DAO::executeQuery("SELECT FROM civicrm_contact WHERE display_name like %1 AND is_opt_out = %2", array(
+$name = 'John Smith'; /* un-trusted data */
+$optedOut = 0;        /* un-trusted data */
+
+$query = "
+  SELECT id
+  FROM civicrm_contact 
+  WHERE 
+    display_name like %1 AND
+    is_opt_out = %2";
+
+$result = CRM_Core_DAO::executeQuery($query, array(
   1 => array('%' . $name . '%', 'String'),
   2 => array($optedOut, 'Integer'),
 ));
@@ -17,7 +27,9 @@ This example ensures that variables are safely escaped before being inserted int
 
 The variable types available for this can be found in [CRM_Utils_Type::validate](https://github.com/civicrm/civicrm-core/blob/60050425316acb3726305d1c34908074cde124c7/CRM/Utils/Type.php#L378). The query engine then applies appropriate escaping for the type.
 
-In some circumstances you may find that a complex query is easier to build by directly escaping values using the `CRM_Utils_Type::escape()` method. It is prefereable to use the form above or the CRM_Utils_SQL_Select format
+## `CRM_Utils_Type::escape` {:#escape}
+
+In some circumstances you may find that a complex query is easier to build by directly escaping values using the `CRM_Utils_Type::escape()` method. It is prefereable to use the form above or the `CRM_Utils_SQL_Select` format
 
 ```php
 $name = CRM_Utils_Type::escape('John Smith', 'String');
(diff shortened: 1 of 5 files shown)


xurizaemon:
I'll take a look! But are you *sure* that was me talked at Ft Collins? I am ... less certain.

docs/standards/sql.md
```diff
@@ -0,0 +1,24 @@
...
+When writing SQL developers should be concious to try to always ensure that all variables that are passed into the SQL are done in a safe and secure maner.
+
+In CiviCRM the best way is to use the inbuild paramatisation tools to do the job. 
+
+```php
+$name = 'John Smith';
+$genderId = 2;
+$result = CRM_Core_DAO::executeQuery("SELECT FROM civicrm_contact WHERE display_name like %1 AND gender = %2", array(
+  1 => array('%' . $name . '%', 'String'),
+  2 => array($genderId, 'Integer'),
```
xurizaemon: For non-SQL reasons, I'm not a fan of CiviCRM's storage of gender data. Maybe `is_opt_out` would be equally effective at demonstrating parameterization?

docs/standards/sql.md
```diff
@@ -0,0 +1,24 @@
+# SQL Coding Standars
+
+When writing SQL developers should be concious to try to always ensure that all variables that are passed into the SQL are done in a safe and secure maner.
+
+In CiviCRM the best way is to use the inbuild paramatisation tools to do the job.
```
xurizaemon: "inbuilt", "parameterization"

docs/standards/sql.md
```diff
@@ -0,0 +1,24 @@
+# SQL Coding Standars
+
+When writing SQL developers should be concious to try to always ensure that all variables that are passed into the SQL are done in a safe and secure maner.
```
xurizaemon: "conscious", "manner"

xurizaemon:
I gave up on my review and PR'd my feedback as seamuslee001/civicrm-dev-docs#1

seanmadsen:
+1 for the practice of submitting feedback as a PR! :slightly_smiling_face: 

I'll look at this too... soon, but not right now.

xurizaemon:
ignore that review - yeah the PR was way better flow!

docs/standards/sql.md
```diff
@@ -0,0 +1,26 @@
...
+In CiviCRM this is done using the inbuilt parameterization tools.
+
+```php
+$name = 'John Smith';
+$optedOut = 0;
+$result = CRM_Core_DAO::executeQuery("SELECT FROM civicrm_contact WHERE display_name like %1 AND is_opt_out = %2", array(
+  1 => array('%' . $name . '%', 'String'),
+  2 => array($optedOut, 'Integer'),
+));
+```
```
totten: This is certainly the material we need to present for common queries.

docs/standards/sql.md
```diff
@@ -0,0 +1,26 @@
...
+
+This example ensures that variables are safely escaped before being inserted into the query. CiviCRM also allows developers to specify the type of variable that should be allowed. In the case of the `%2` ($optedOut) parameter, only an *Integer* input will be permitted.
+
+The variable types available for this can be found in [CRM_Utils_Type::validate](https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/Type.php#L378). The query engine then applies appropriate escaping for the type.
+
+In some circumstances you may find that a complex query is easier to build by directly escaping values using the `CRM_Utils_Type::escape()` method. It's preferable to use the form above.
```
totten: For complex query building, I think that we're inviting if we suggest that people use DAO. I'd rather point towards:

https://github.com/civicrm/civicrm-core/blob/master/CRM/Utils/SQL/Select.php#L33

totten: Since the topic of complex query-building came up, I'm ready to cave on the purity of `CRM_Utils_SQL_Select`... https://github.com/civicrm/civicrm-core/pull/10686

seamuslee001: @totten i have attempted to put some docs in let me know what you think.

seanmadsen:
@seamuslee001 thanks for this work. This stuff is really great! I made some small improvements: 

* Formatting
    * It's worth pointing out that I changed the html table to separate sections with headings because this means that we can use markdown for links and inline code. I'll admit that I think the table you created was slightly easier to read. My change makes a small sacrifice in readability, but comes with some gains in readability (for inline code) and comes with gains in terms of maintainability (since the doc source is cleaner). Also, when we use markdown for links, it becomes easier for us to ensure that the guides don't have any broken internal links. Let me know if you have opinions about this change. I'm open to other approaches. 
* Added the pages you created to the nav menu `mkdocs.yml`. 
* Added a general "Code standards" page
* Added a redirect from the wiki's "PHP standards" page so we can officially mark that wiki page as having been migrated.

seanmadsen:
Given @totten's [concerns](https://github.com/civicrm/civicrm-dev-docs/pull/222#discussion_r127848710) about suggesting that people use the DAO, we are now waiting on @totten to either: 

* give the go-ahead for merging this PR as-is (with additions from @seamuslee001 about `CRM_Utils_SQL_Select`)
* OR suggest an alternative

I will merge this PR in 5 days if there are no further comments. I think that right now this content is "better in than out". Reading this page was actually the first time I had ever heard about `CRM_Utils_SQL_Select`!

totten:
I agree. This is a good improvement to the docs, and it's better in than out.

seamuslee001:
@seanmadsen yep your changes are good mate, I just flowed on from the previous, I think it would be good if we could get a redirect @totten @mlutfy https://wiki.civicrm.org/confluence/display/CRMDOC/PHP+Code+and+Inline+Documentation to https://docs.civicrm.org/dev/en/latest/standards/php/ now

seanmadsen:
@seamuslee001 the redirect happens automatically when we add it to the book's redirect file [as I did in this commit](https://github.com/civicrm/civicrm-dev-docs/pull/222/commits/bb435f2c24a8a6909d2bf134a6cde3c3dbff70f7), but it's not instantaneous. Apparently there's roughly a 10 minute delay.

seamuslee001:
@seanmadsen ahhh cool :-)

seamuslee001:
@JoeMurray Joe i think an example would be probably which is something i have seen in one of our custom modules 

A developer realises they are going to be using the same base block of sql in multiple places so rather than writing it in each they generic block is put into a helper function where it builds joins + some of the where perhaps but then is fed out to the out functions where its put together with some more sql probably more where clauses and then run through the CRM_Core_DAO::executeQuery, So in this case in that helper function it would make sense to use the CRM_Utils_Escape:; if your only wanting to send back an SQL string. 

I can also point to this example in the current codebase as a situation https://github.com/civicrm/civicrm-core/blob/master/CRM/Mailing/Event/BAO/TrackableURLOpen.php#L58 i think Chris or Tim talked with Mathieu when coming up with the appropriate escaping strategy in Colarado last year.

Issue #193, opened 2017-06-17
Find a place for "Upgrade Tests" tutorial


seanmadsen:
TL;DR: Grab the content from [Upgrade Tests](https://github.com/civicrm/civicrm-buildkit/blob/d20229ac9f63aaf183687157a5e655c953b81f5e/doc/daily-coding.md#upgrade-tests) and find a home for it in the Dev Guide.

Longer story: I'm migrating docs content out of the [buildkit repo](https://github.com/civicrm/civicrm-buildkit) and into the Dev Guide and as part of this process I'm deleting that [file](https://github.com/civicrm/civicrm-buildkit/blob/d20229ac9f63aaf183687157a5e655c953b81f5e/doc/daily-coding.md) which contains a mix of content that needs some work to be well-incorporated into the Dev Guide. I'm using issues to track the process of this migration, section-by-section, since I think the content will likely need to go to multiple places.

seanmadsen:
@totten suggested [here](https://github.com/civicrm/civicrm-dev-docs/pull/223#discussion_r128615202) that perhaps we should put this "upgrade tests" content in `docs/framework/upgrade.md`

seamuslee001:
@seanmadsen i think this can be closed now?

seanmadsen:
ahh yes. Thanks :slightly_smiling_face:

PR #223, opened 2017-07-20
Add in upgrade reference documents

seamuslee001:


seanmadsen:
@seamuslee001 this is awesome! 

You are migrating stuff from the wiki and this is so so so good! 

If you want to keep doing this for more wiki pages, here are some tips (entirely optional) that would make the process go a little faster on my end: 

* Use a text editor with a spell-checker
* Preview locally with `mkdocs serve` (instructions here) (I'm only guessing that you're not doing this because your latest two PRs failed to build with mkdocs because of preceding forward slashes in the paths in `mkdocs.yml`)
* Use the [webpage2md](https://github.com/seanmadsen/webpage2md) tool to import content from the wiki
* Make separate commits for content imported with the tool vs your changes on top of that

Again, these tips are optional &mdash; overall, your PRs are splendid exactly as you're making them!

seanmadsen:
I've made some improvements to this PR for doc formatting, but I'm going to leave this PR open for 2 more days in case @totten (or anyone else) want to review it for content accuracy.

docs/framework/upgrade.md
```diff
@@ -81,5 +81,4 @@ All Smarty tags are evaluated before the sql is run. Commonly used Smarty variab
 
 Testing When testing upgrades, it's important to run the upgrades against several different databases (representing different versions of CiviCRM and different upgrade paths). The tool "civicrm-upgrade-test" can automate this process. For instructions:
 
-* If you are using buildkit you can use the integrated [civibuild upgrade-test](https://github.com/civicrm/civicrm-buildkit/blob/master/README.md#daily-coding-upgrade-testing) command.
```
totten: Aww, did we lose the tutorial for the upgrade-testing workflow? Maybe it should migrated into this doc?

https://github.com/civicrm/civicrm-buildkit/blob/37abb2e6dd9790ffc016377dc2e20c0965e2953d/doc/daily-coding.md#upgrade-tests

seanmadsen: Here's where it "went" https://github.com/civicrm/civicrm-dev-docs/issues/193

docs/framework/upgrade.md
```diff
@@ -0,0 +1,84 @@
+# Upgrade Reference
+
+!!! note "Scope"
+    This document discusses upgrade conventions used in CiviCRM core code. For extensions and modules which integrate with CiviCRM, upgrade conventions are different. See [Create a Module Extension](/extensions/civix.md) for more discussion of the module-focused upgrader.
+
+## Introduction
+
+Within CiviCRM Core code the upgrader classes handle making any necessary database changes either through PHP or SQL code, printing out any messages before or after the upgrade happens. Queuing up tasks for each version upgrade.
```
totten: * As an "introduction", this is probably better than the old one - ie it quickly summarizes *what* happens rather than digging into *how*.
 * The last sentence is not a sentence. Kill it?
 * It is useful to break-down the mechanics of the process. I guess the question is whether material from [Introduction](https://wiki.civicrm.org/confluence/display/CRMDOC/Upgrade+Reference#UpgradeReference-Introduction) and [Order of Upgrade Steps](https://wiki.civicrm.org/confluence/display/CRMDOC/Upgrade+Reference#UpgradeReference-OrderofUpgradeSteps) are duplicates?

docs/framework/upgrade.md
totten: Got it, thanks. :)  Seems like it's not a blocker for #223.

docs/framework/upgrade.md
seamuslee001: I think its useful having the order of the upgrade steps so people are aware of how things do get run and when, Like the thing about verify DB state only happening on major upgrades is interesting

seamuslee001:
@seanmadsen what do you think of Tim's comment also what do you think of the state of this PR

seanmadsen:
Thanks for the ping, @seamuslee001 

Ok, I added a commit to remove the incomplete sentence, which addresses @totten's 2nd point [above](#discussion_r128618161). 

I don't completely have my head around Tim's 3rd point, but I gather that Tim and Seamus have slightly different opinions there. I also gather that the issue is pretty minor. So I'm merging as-is. We can sort out any remaining improvements in subsequent PRs.

PR #224, opened 2017-07-24
Improve buildkit installation docs

seanmadsen:
I installed buildkit on a new machine today and got hung up on some things. I hope these docs improvements will help others avoid the pitfalls I encountered today. 

@xurizaemon @seamuslee001 care to take a look at this?

### DIFF
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -79,22 +79,6 @@ npm update
 npm install fs-extra
 ```
 
-## Applying a patch
-
-Using buildkit, you can create a CiviCRM environment with the PR applied.
-
-For example:
-
-```bash
-civibuild create dmaster \
-  --url http://localhost:8001 \
-  --patch https://github.com/civicrm/civicrm-core/pull/8494 \
-  --admin-pass s3cr3t
-```
-
-This will create a test environment with the Drupal, CiviCRM master branch
-and the patch in PR 8494. More detailed information is in the
-[Civibuild documentation](https://github.com/civicrm/civicrm-buildkit/blob/master/doc/civibuild.md)
 
 ## Post-install configuration {:#configuring}
 
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -96,7 +96,9 @@ This will create a test environment with the Drupal, CiviCRM master branch
 and the patch in PR 8494. More detailed information is in the
 [Civibuild documentation](https://github.com/civicrm/civicrm-buildkit/blob/master/doc/civibuild.md)
 
-## Configuring buildkit after installation {:#configuring}
+## Post-install configuration {:#configuring}
+
+### Configuring your path {:#path}
 
 !!! note "Not needed for Vagrant/Docker installations"
     If you set up buildkit using Vagrant or Docker, then you don't need to perform the configuration steps listed here.
@@ -134,6 +136,41 @@ If you want to ensure that the buildkit CLI tools are always available, then:
 
     Each time you open a new terminal while working on Civi development, you would need to re-run the `export` command.
 
+
+### Configuring `amp` {:#amp-config}
+
+!!! tip
+    Login as a non-`root` user who has `sudo` permission. This will ensure that new files are owned by a regular user, and (if necessary) it enables `civibuild` to restart Apache or edit `/etc/hosts`.
+
+The first build requires only a few commands.  However, these are also the
+hardest commands -- you need to provide detailed information about the
+Apache/MySQL/PHP systems, and you may need to try them a few times.
+
+Configure `amp` with details of your Apache/MySQL environment.  Pay close
+attention to the instructions.  They may involve adding a line to your
+Apache configuration file.
+
+```
+$ amp config
+```
+
+Test that `amp` has full and correct information about Apache/MySQL.
+
+```
+$ amp test
+```
+
+!!! note
+    You may need to alternately restart httpd, re-run `amp config`, and/or re-run `amp test` a few times.
+
+Create a new build using Drupal and the CiviCRM `master` branch.
+The command will print out URLs and credentials for accessing the website.
+
+```
+$ civibuild create dmaster --url http://dmaster.localhost --admin-pass s3cr3t
+```
+
+Once you have a working build of `dmaster`, you can continue working with `civibuild` to create different builds as described below.
 
 
 ## Upgrading buildkit {:#upgrading}
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -112,63 +112,60 @@ If you want to ensure that the buildkit CLI tools are always available, then:
 
 Buildkit provides a tool called `amp` which [civibuild](/tools/civibuild.md) uses when it needs to set up a new site. Before you can use `civibuild`, need to configure `amp` by telling it a bit about your system (e.g. what webserver you're using). 
 
-#### Interactive config
+1. Run the interactive configuration tool.
 
-```
-$ amp config
-```
-
-!!! tip "tips"
-    * Run this as a non-`root` user who has `sudo` permission. This will ensure that new files are owned by a regular user, and (if necessary) it enables `civibuild` to restart your webserver and edit `/etc/hosts`.
-    * Pay close attention to any instructions given in the output of this command.  They may involve adding a line to your Apache or nginx configuration file.
-    * To check which version of apache you have, run `apachectl -v`
-
-#### Testing amp's configuration {:#amp-test} 
-
-Test that `amp` is correctly configured.
-
-```
-$ amp test
-```
-
-!!! failure "Troubleshooting errors from `amp test`"
+    ```
+    $ amp config
+    ```
     
-    * Try manually adding settings to your webserver, as described below.
-    * Re-run `amp config`.
+    !!! tip "tips"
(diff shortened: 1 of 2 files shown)


seanmadsen:
BTW, one other thing I got hung up on was thinking that I had some sort of problem with MySQL packages when actually the real culprit was that I never ran `amp config`. I got turned around because of this warning message during the initial installation process: 

> "Warning: This system will not support php-mysql extension"

I made [this PR](https://github.com/civicrm/civicrm-buildkit/pull/339) which attempts to mitigate this pitfall by removing that warning message.

docs/tools/buildkit.md
```diff
@@ -135,6 +108,84 @@ If you want to ensure that the buildkit CLI tools are always available, then:
...
+
+#### Interactive config
+
+```
+$ amp config
+```
+
+!!! tip "tips"
+    * Run this as a non-`root` user who has `sudo` permission. This will ensure that new files are owned by a regular user, and (if necessary) it enables `civibuild` to restart Apache or edit `/etc/hosts`.
+    * Pay close attention to any instructions given in the output of this command.  They may involve adding a line to your Apache configuration file.
```
seamuslee001: Should also mention nginx perhaps change to "Pay close attention to any instructions given in the output of this command. They may involving adding a line to your web-server configuration (apache or nginx)

docs/tools/buildkit.md
```diff
@@ -123,38 +110,81 @@ If you want to ensure that the buildkit CLI tools are always available, then:
...
+    
+        Create a new file `/etc/apache2/conf.d/buildkit.conf` with the following contents:
+    
+        ```
+        IncludeOptional <amp-installation>/apache.d/*conf
+        ```
+
+    * For nginx:
+    
+        Create a new file `/etc/nginx/conf.d/buildkit.conf` with the following contents:
```
seanmadsen: Did I get this path correct @xurizaemon ? `/etc/nginx/conf.d/buildkit.conf`

docs/tools/buildkit.md
seanmadsen: Fixed in bb27c6e

docs/tools/buildkit.md
xurizaemon: "It depends" ... For Debian, the path above is correct assuming they're using the Debian NginX packages yes. But maybe not all ... eg Debian and CentOS differ on Apache (`/etc/apache2` vs `/etc/httpd`), for my OSX with homebrew it's `/usr/local/etc/nginx/servers/buildkit.conf`, and that's before you get into the madness of things like CPanel and their ilk, or people installing to `/usr/local/nginx`...

seanmadsen: I see. Do you think it's okay to merge this as-is and assume that people will figure it out? Or do you have something better to propose here?

docs/tools/buildkit.md
```diff
@@ -135,6 +108,84 @@ If you want to ensure that the buildkit CLI tools are always available, then:
...
+
+Test that `amp` is correctly configured.
+
+```
+$ amp test
+```
+
+!!! failure "Troubleshooting errors from `amp test`"
+    
+    * Try manually adding settings to your sebserver, as described below.
```
tschuettler: webserver

docs/tools/buildkit.md
totten: IMHO, you should only document `amp` with `apache`. `nginx` doesn't read `.htaccess`, and `amp` doesn't have enough plumbing to handle bespoke `nginx` config files. Basically, `civibuild`/`amp` needs a way to pass through per-application templates for the nginx config.

Without that, it's easier to (a) just tell people to use apache or (b) turn off amp's httpd integration and do that stuff manually.

docs/tools/buildkit.md
```diff
@@ -135,6 +108,84 @@ If you want to ensure that the buildkit CLI tools are always available, then:
...
+    * Re-run `amp config`.
+
+#### Manually adding settings to your webserver {:#amp-webserver}
+
+During one of the steps in the `amp config` process, `amp` will *attempt to* alter the system-wide settings for your local webserver (apache or nginx). On some platforms, `amp` is not able to perform the necessary configuration and you must do it manually by following these steps:
```
totten: This seems like a misunderstanding -- `amp` makes *no attempt* to modify the system-wide httpd configuration. Rather, it *tells the user to do it* (and tries to give some pointers in the right direction). Of course, half the users ignore those instructions, so it's a recurring pain-point. 

Random:
 * Guess: there's confusion because the vagrant and docker builds do that part automatically. 
 * But as far as `amp config` and manual setup are concerned, these aren't automated.
 * It would be patch-welcome for `amp config` to put up a better prompt.
 * It would be patch-welcome for `amp config` to try to apply the needful.
 * It would be patch-welcome for `civi-download-tools --full` to try to apply the needful.

seamuslee001: Agree that it doesn't alter apache or nginx it does modify the hostfile which is http related, I suppose the most sensible thing i can think amp doing would be is to write out the include as a new file in the right directory. But as Chris pointed out finding that correct directory maybe tricky

totten: Yup. Currently the `amp config` wizard just displays a handful of possibilities and moves on to the next step -- https://github.com/amp-cli/amp/blob/master/src/Amp/Command/ConfigCommand.php#L139

In an ideal world, that script would stop and ask the user what to do. Depending on what config files/directories are around, it would present a menu like:

```
[1] Include amp in /etc/apache2/httpd.conf
[2] Include amp in /etc/apache2/conf.d/amp.conf
[3] Include amp in /opt/xampp/etc/httpd.conf
[4] Skip this step. I'll configure it manually.
```

If they choose 1-3, it would call `sudo` and try to update the selected file.

This way, it's harder for a user to ignore the step, but we still rely on their judgment for where/how to make the change.

docs/tools/buildkit.md
seanmadsen: Thanks @tschuettler. Fixed in 3485d726047aa8ba7cd61c6b911b7ee516cd52ff

docs/tools/buildkit.md
seanmadsen: @totten indeed I was misunderstanding the expected behavior of `amp config`. I have made changes in 0015e66 to clarify that, indeed, adding the settings to the webserver is mandatory.

seanmadsen:
I have made adjustments to this PR which I think addresses feedback from @totten @tschuettler and @seamuslee001. Will merge in 2 days if no further comment. 

I recognize that the issue of "whether (and if so, how much) to document `amp` vis-à-vis nginx" is still somewhat outstanding in this PR. @totten seems to be arguing that we remove the references to nginx. @seamuslee001 has argued for more references to nginx in this same PR. So I'm a bit torn as to how to address this feedback, given that (personally) I don't know much about nginx. As such, I lean towards merging this content as-is and addressing this issue in a separate PR. My eagerness to merge is somewhat higher in this PR than in most because I know that installing buildkit has been super challenging for lots of people, so I want to get this published, even if it's not perfect.

seamuslee001:
Agree with you Sean, my argument for including nginx is that it seems to me from the line that Tim linked to earlier we are actively potentially supporting nginx for httpd as also evidenced by https://github.com/amp-cli/amp/blob/master/src/Amp/Command/ConfigCommand.php#L329 so to me it feels like we should document how to potentially set that up but noting it is still WIP for nginx but more solid for apache and apache is included in the download tools command

totten:
@seamuslee001 I've been tempted to rip that code out. ;) In my mind, it's harmful/misleading because there is no good way to setup `nginx`, just a choice between "not working at all" (default) and "not working half the time" (maybe achievable, if you customize the vhost template and configure it to match a specific build type like `drupal-demo` or `wp-demo`). 

IMHO, it's really a waste of time to document a broken design. If you want to put energy into `nginx` support, then try these:

 * Update `amp` (I've implemented; needs reviewer)
   * As discussed in https://github.com/amp-cli/amp/issues/21#issuecomment-96271971, `amp create` should figure out the nginx template using a dot-file (eg `MYBUILD/.amp/config.yml` or `MYBUILD/.amp/nginx-vhost.php`).
   * I've posted a PR to do this: https://github.com/amp-cli/amp/pull/47 (example build `http://think.hm/tmp/amp.phar-pr47`) Let me know if itworks.
 * Update `civicrm-buildkit` (needs implementer and reviewer)
   * Add files like `app/config/*/nginx-vhost.php`. (These would like https://github.com/amp-cli/amp/blob/master/src/Amp/views/nginx-vhost.php but be tuned to the needs of each CMS+Civi config.)
   * Patch `civibuild`: it should either set an env var (`NGINX_VHOST_TPL=$BLDDIR/app/config/*/nginx-vhost.php`); or the `install.sh` scripts should copy over the file to `CMS_ROOT/.amp/nginx-vhost.php`.

seamuslee001:
@totten i would be ok with removing the documentation if we also removed the nginx option (could we comment it out) until we get buildkit ready for nginx (for some reason i thought @xurizaemon had done exactly that), Just figured that if it was a live option in the app we should have documentation for it if its really a half arsed set up we should pull it out of the app until its properly sorted IMHO

seanmadsen:
I added a "caution" admonition which recommends that users choose Apache over nginx. I will stick to my deadline of merging this PR tomorrow if there are no further concerns. Then, if we get more support nginx we can remove this word of caution. Or if we drop support for nginx altogether we can update the docs to reflect.

PR #227, opened 2017-07-25
CRM-20926 - hook_civicrm_{alter,xml}Menu - Fill in docs. Add PHPIDS docs.

totten:
This fills in some missing documentation for the menu hooks, and it adds new documentation for the
`ids_arguments` defined by CRM-20926.

Testing tips: 
 * If you'd like test a particular form, you can try opening your form in a browser and
   passing a suspicious input (as in https://gist.github.com/totten/27cc1bde206fe17371c8a6d7cb361e49).
 * If you'd like to inspect the exact ratings produced by PHPIDS for the inputs, see
   the throw-away script `check-ids-value.php` (also in https://gist.github.com/totten/27cc1bde206fe17371c8a6d7cb361e49).

### DIFF
--- a/docs/hooks/hook_civicrm_alterMenu.md
+++ b/docs/hooks/hook_civicrm_alterMenu.md
@@ -4,10 +4,6 @@
 
 This hook is called when building CiviCRM's list of HTTP routes and should be used when you want to register custom paths or URLS.
 
-## Notes
-
-You will need to visit `/civicrm/menu/rebuild?reset=1` to pick up your additions.
-
 !!! note "Comparison of Related Hooks"
     This is one of three related hooks. The hooks:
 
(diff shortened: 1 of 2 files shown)


seanmadsen:
@totten I'm still working on absorbing/learning all the info in this PR (as well as https://github.com/civicrm/civicrm-core/pull/10709) in pursuit of providing a thorough review of both. 

**Question**: Is there are reason you opted to put [all the info about menu XML tags](https://github.com/totten/civicrm-dev-docs/blob/47aec12d9d5581cacdc3f0f6cc7860aef9953b90/docs/hooks/hook_civicrm_xmlMenu.md#xml-common) in the hook page? I would expect that info to be in a page alongside [framework/schema-definition](https://docs.civicrm.org/dev/en/latest/framework/schema-definition/). If you agree, I don't think this needs to be a blocker to this PR. We can move it later, (and also I think it'd be nice to put the menu XML docs into the same format as the schema definition docs). Right now I'm just trying to better understand these concepts, since all the menu stuff and IDS stuff is totally new to me. Hence my question of why you put it in the hook page.

totten:
@seanmadsen I basically agree with everything you've said. :) The "Framework => Schema Definition" is a nicer a layout, reworking in toward that would be a good change, but it's not a  blocker.

seanmadsen:
These docs look good 

Ready to merge this when https://github.com/civicrm/civicrm-core/pull/10709 is merged

PR #229, opened 2017-07-26
Update phpunit.md to document PHPUNIT_TESTS functionality

mattwire:


seanmadsen:
I changed the "warning" admonition to a "note" admonition. Is this okay? If it really is a *warning* I think it'd be nice to explain a little more about why.

mattwire:
@seanmadsen That's perfect.  It's the first time I've committed to the dev docs and I didn't make time to find out what the choices were :-)

seanmadsen:
@mattwire Awesome. For future reference, here are [the choices](https://docs.civicrm.org/dev/en/master/documentation/markdown/#admonitions). But also it's great that you're not letting those details get in the way of your ability to contribute to docs!

totten:
Hey, this env var is unique to Civi, so 👍 for documenting it.

Full disclosure: for a critical a reader who knows `phpunit`, it might feel a little quirky -- because `phpunit` accepts some CLI arguments for selecting the target tests (e.g. `--group` or `--filter` or the file-name). And it's quite well-and-good to use those normal options.

The `EnvTests` arose to address a fairly weird niche: https://forum.civicrm.org/index.php?topic=33795.0

PR #237, opened 2017-08-01
Changed $objectName to $objectType for consistency

twomice:
Examples and explanations on this page use $objectType, so standardizing on that.

### DIFF
--- a/docs/hooks/hook_civicrm_searchTasks.md
+++ b/docs/hooks/hook_civicrm_searchTasks.md
@@ -7,11 +7,11 @@ search, allowing you to inject additional actions or to remove existing actions.
 
 ## Definition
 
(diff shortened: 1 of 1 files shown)


seanmadsen:
[Looks like](https://github.com/civicrm/civicrm-core/blob/5d9bcf011d4705c18c78b43ce82becae6474d51d/CRM/Utils/Hook.php#L904) the function signature also uses `$objectType`, so this PR looks good :+1:

totten:
Nice catch. It's particularly good/important that the docs now match the variable names from `CRM_Utils_Hook` -- for [hook listeners using an $event object](https://docs.civicrm.org/dev/en/master/hooks/setup/symfony/), the parameter names are a firm part of the contract.

PR #241, opened 2017-08-04
Add security note about AngularJS in Smarty

seanmadsen:
I think this PR is pretty straightforward. Will merge in 5 days if no one provides feedback. 

@seamuslee001 you might be interested in this. I was thinking that potentially we could have an "AngularJS" section within your security page, perhaps just with this content in it. If you think that's a good idea, then we could move this content over when you make your page, and link to it from where I've added it now.

### DIFF
--- a/docs/framework/angular/loader.md
+++ b/docs/framework/angular/loader.md
@@ -160,6 +160,24 @@ in the Smarty template:
 </div>
 ```
 
+!!! caution "Security note"
+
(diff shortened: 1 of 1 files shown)


totten:
Truth!

Would anyone actually consider doing that? 

(disclosure: after talking to Frank about Volunteer, we came up with a change which would make it even less likely: https://github.com/civicrm/civicrm-core/pull/10783 )

seamuslee001:
This looks good mate +1

PR #243, opened 2017-08-04
Rename "Schema definition" to "Database Schema"

totten:
1. There are multiple XML schemas -- eg for database definition, for routing, for CiviCase workflows, for Symfony services. Specify which one.
 2. The other titles in the menu use "Title Case".


totten:
Ooh, I may have interpreted "Schema" as "XML schema", but this document is the "XML schema" which defines the "database schema". I guess it could have been intended as "schema" in the second sense (and I misinterpreted as the first sense).

seanmadsen:
I think this is pretty minor and I'm fine either way, but I do agree with your second comment more than your first (which means I support leaving the code as-is and closing this PR). 

If I saw a page called "Database Schema", I would think it would describe all the tables and fields defined for CiviCRM. This sort of page is actually something that people [appear to want](https://civicrm.stackexchange.com/questions/19704/er-diagram-for-the-latest-civicrm-foreign-key-in-structure-test-cases-plans), but the current page doesn't provide any of that info (which is good, IMHO).

seamuslee001:
What about Database XML Structure or something, I agree with Tim about just saying XML Schema is not right, but i agree with Sean that Database Schema would lead people to think its a ERD or similar

seanmadsen:
Closing in favor of: https://github.com/civicrm/civicrm-dev-docs/issues/256

PR #245, opened 2017-08-05
Security related documentation

seamuslee001:
@totten @seanmadsen open for some review, I think this meets the balance we were looking for but i like the concept of a Security chapter for reporting bug + ACL and permissions but then in the coding standards have a thing on doing this all correctly.


docs/standards/secure.md
```diff
@@ -0,0 +1,49 @@
...
+<a  href="{$item.url}" title="{$item.title|escape:'html'}">
+```
+
+This will ensure that the variable title within the item key when generating a list of recently viewed items won't have any Cross Site Scripting as it will be escaped for use within HTML. For more information on the types of escaping you can do with Smarty see the [Smarty Documentation](https://www.smarty.net/docsv2/en/language.modifier.escape)
+
+However sometimes to escape on output you need to ensure that because of the complex nature of the variable that the variable is properly escaped when passed to Smarty. For example, when building a json encoded blob of data for use in an contribution form it was necessary to escape before passing onto the Smarty Template.
+
+```php
+$form->assign('submittedOnBehalfInfo', json_encode(str_replace('"', '\"', $form->_submitValues['onbehalf']), JSON_HEX_APOS));
+```
```
seanmadsen: @seamuslee001 I'm a little confused about this code block and worried that it might be a bad example. But maybe I'm missing something. Here's my thinking: 

* My understanding is that `str_replace('"', '\"', $untrustedData)` is bad practice because an attacker can pass a string like `foo\" bar` and effectively "escape the escape". So reading that threw up a red flag for me.
* But then I see the `json_encode` function and think *"oh, ok, that's good"*. I don't really understand why `str_relace` is there. Maybe we could use a simpler example here? Or explain more?

docs/standards/secure.md
```diff
@@ -0,0 +1,49 @@
...
+
+For AngularJS templates, developers should consult the AngularJS [$sanitize documentation](https://docs.angularjs.org/api/ngSanitize/service/$sanitize).
```
seanmadsen: I'm not positive about this, but I *think* that with our recommended (and documented) practices of fetching data through the AJAX API within AngularJS, there's actually no need to use `$sanitize`. Am I right about that, @totten @colemanw? This is not super fresh in my memory, but I recall reading about `$sanitize` and thinking "Oh I don't need to know how to use that". If this is the case, I think it'd be best to remove this line which mentions it, or explain more about the special circumstances that warrant using `$sanitize`.

docs/standards/secure.md
seamuslee001: I'm not sure of a good current example from the code apart from that where we are doing something this complex. This one was from the most recent round of security fixes. @totten may have ideas

docs/standards/secure.md
seamuslee001: I was more thinking of the variables passed to it in boot e.g. through CRM_Mailing_Info or whatever the developer may do, But i'm happy to get word from those two if we don't need it

mkdocs.yml
```diff
@@ -73,6 +73,11 @@ pages:
   # CiviMail: /reference/civimail.md
   # CiviReport: /reference/civireport.md
   # Payment Processing: /reference/payment.md
+- Secure Coding:
```
seamuslee001: @seanmadsen the title here may warrant a change to diff from New Line 219

seanmadsen:
@seamuslee001 Want to take a look over these additions from me?

docs/security/inputs.md
```diff
@@ -0,0 +1,51 @@
...
+```php
+$angPage = CRM_Utils_Request::retrieve('angPage', 'String', $this);
+if (!preg_match(':^[a-zA-Z0-9\-_/]+$:', $angPage)) {
+  CRM_Core_Error::fatal('Malformed return URL');
+}
+```
+
+## `POST` parameters
+
+When accepting `POST` parameters through forms, it's important to validate the data as
```
seamuslee001: @seanmadsen this looks odd

seanmadsen: Ha! Yes. Thank you for catching that. Fixed (sort of) in e96ce38628eb7c08bbc0976eed6a81ab1de27336 although I don't actually know a whole lot about the best ways to do form validation, so I'm not sure what else to write in the section (which I guess is why I forgot to finish it!). If you have a good sense of these concepts, maybe you could add some content here? Although I think we could merge this as-is, even though it's not very complete at the moment. Then add more content later.

seamuslee001:
I think the changes make sense @seanmadsen i would like @totten to have a look over them too,

seanmadsen:
I'll merge this on 2017-08-17 if we don't hear from anyone else.

seanmadsen:
@seamuslee001 your changes from today look good :white_check_mark: 

Still on-track to merge on 2017-08-17 if no further comments.

MegaphoneJon:
I reviewed this today.

There is overall a great addition, and there's nothing I see that would block a merge.  If time permits, then some time down the line, I might consider rewriting the intro (use it to introduce the other pages a bit more) and there's some small punctuation things, but I think we merge now and improve later.

seamuslee001:
thanks @MegaphoneJon :-)

seanmadsen:
Merging now since we got a review from @MegaphoneJon :tada:

PR #252, opened 2017-08-07
Update phpunit.md to reflect discussions on chat.civicrm.org

mattwire:
Update based on discussions with @mickadoo @totten


totten:
Cool :)

PR #253, opened 2017-08-07
Port Resources Refecnce from WIki

seamuslee001:
ping @totten @seanmadsen 

I did some minor re-writes but mostly copied from the old wiki page pls check


seanmadsen:
Looks good. I made some small tweaks. Thanks so much for working on this wiki migration stuff, @seamuslee001 !!

PR #254, opened 2017-08-08
Adding in more detailed reference -> to PHPStorm set up (with screens…

KarinG:
…hots!)

and reminding everyone to always use buildkit


seanmadsen:
Thanks Karin! I made a small formatting adjustment, but this looks good :slightly_smiling_face:

PR #264, opened 2017-08-11
Migrate "Extension Reference" wiki page

seanmadsen:
Reviews welcome. Will merge in 5 days if no comments.

This page migration really ballooned into way bigger of a project than I wanted. I ended up researching quite a bit about [XSD](https://en.wikipedia.org/wiki/XML_Schema_(W3C)) and [RELAX NG](https://en.wikipedia.org/wiki/RELAX_NG) in hopes that I could generate some good-looking docs from an `.xsd` or `.rng` file. But unfortunately the available tools for turning schema definition into doc are not good. In the process of this research, however, I did learn a bit more about how people typically describe XML schemata. Even though this page was a lot of work to put together, I think this format is ultimately a bit superior to the [database schema definition](https://docs.civicrm.org/dev/en/latest/framework/schema-definition/) page. It is a bummer that the pages are in different formats. I wrestled with this a bit. Anyway, there's lots more I could say about all this, but like I said, I've already spent too much time on this one. Phew!

### DIFF
--- a/docs/extensions/info-xml.md
+++ b/docs/extensions/info-xml.md
@@ -1,330 +1,356 @@
 # info.xml
 
-## Introduction
+Every CiviCRM [extension](/extensions/index.md) must have an `info.xml` file within it to provide information about the extension. This page is a reference for the schema of the `info.xml` file.
 
-A ***native CiviCRM extension*** is a feature provided by a third-party
-developer which can be installed on CiviCRM. To support automated
-distribution and installation, an extension must be packaged according
-to a particular specification. This page documents the technical
-structure of an extension. For a proper introduction to developing
-extensions, see [Create an Extension](/extensions/index.md).
+Typically, you'll begin by running [civix generate:module](/extensions/civix.md#generate-module) and `civix` will create a basic `info.xml` file for you. Use this page to learn how to customize that file.
+
+## Example
+
+Here is an example of a full `info.xml` file from [CiviVolunteer](https://github.com/civicrm/org.civicrm.volunteer/blob/master/info.xml).
+
+```xml
+<?xml version="1.0"?>
+<extension key="org.civicrm.volunteer" type="module">
+  <file>volunteer</file>
+  <name>CiviVolunteer</name>
+  <description>The CiviVolunteer extension provides tools for signing up, managing, and tracking volunteers.</description>
+  <license>AGPL-3.0</license>
+  <maintainer>
+    <author>Ginkgo Street Labs; CiviCRM, LLC; and the CiviCRM community</author>
+    <email>inquire@ginkgostreet.com</email>
+  </maintainer>
+  <releaseDate>2016-12-06</releaseDate>
+  <version>4.6-2.2.1</version>
+  <develStage>stable</develStage>
+  <compatibility>
+    <ver>4.6</ver>
+    <ver>4.7</ver>
+  </compatibility>
+  <comments>
+    Developed by Ginkgo Street Labs and CiviCRM, LLC with contributions from the community. Special thanks to Friends of Georgia State Parks &amp; Historic Sites for funding the initial release, and to The Manhattan Neighborhood Network for funding the 1.4 release.
+  </comments>
+  <civix>
+    <namespace>CRM/Volunteer</namespace>
+  </civix>
+  <urls>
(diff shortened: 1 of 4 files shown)


docs/extensions/info-xml.md
```diff
@@ -0,0 +1,356 @@
...
+### `<comments>` {:#comments}
+
+* Containing element: [`<extension>`](#extension)
+* Description: A long description of the extension that will display to the user before installation.
+* Contains: text
+
+### `<compatibility>` {:#compatibility}
+
+* Containing element: [`<extension>`](#extension)
+* Description: specifies the versions of CiviCRM which which this extension is compatible
```
tschuettler: with which

docs/extensions/info-xml.md
```diff
@@ -0,0 +1,356 @@
...
+
+* Containing element: [`<extension>`](#extension)
+* Description: Stores all the urls that you think are appropriate and necessary for users to find out more about what your extension does, additional documentation, etc.
+* Contains: elements
+
+Elements acceptable within `<urls>`
+
+| Element | Acceptable instances |
+| -- | -- |
+| [`<url>`](#url) | 1 |
```
tschuettler: | [`<url>`](#url) | 1+ |

seanmadsen:
@tschuettler thank you for catching the missing doc about `<civix>` and `<namespace>`. These were not part of the doc in the wiki page which is why I missed them :slightly_smiling_face: I added that in ebe2d166740bd911967a4ec6cff87acbc23020e2. 

@tschuettler or @totten did I get this correct?

> `<namespace>` - The PHP namespace that `civix` uses when generating code 

Or is there a better way to describe that?

docs/extensions/info-xml.md
seanmadsen: Fixed in e3c368df66cba1c1b93dca6d47b29e4ab39cbc08

docs/extensions/info-xml.md
seanmadsen: Fixed in 74e35acedfac3ef79674e72a1e05f7c46610d8bd

totten:
@seanmadsen yup, that's correct. The `<namespace>` is *only* used when running `civix` to generate code. (It's not used at runtime, and you're free to manually write code that follows a different convention.)

seamuslee001:
@totten how does the name space of extension get registered for say PSR4 style loading? / what sets the namespace in that situation

totten:
@seamuslee001 there's an example of PSR-4 class-loading in https://github.com/civicrm/api4/blob/master/info.xml#L23 . If you're implying that we should document those tags, then I agree. :)

Note that generated code produced by `civix` still uses the PEAR/`include_path` convention for class naming/loading. I've been hesitant to change `civix` to namespaces/PSR-4 because (a) it'd take half a day just to update the templates, (b) it'd break compatibility with 4.6, and (c) I'm not sure that all subsystems are actually compatible with namespaces yet. Most likely the transition would involve some kind of opt-in (e.g. `<civix><psr4>true</psr4></civix>` would mean "code generated for this extension will use prettier namespaces").

seamuslee001:
@totten true and yes i think its worth documenting it especially for anyone thinking of D8 extensions if we are heading towards a composer build for D8

seanmadsen:
@totten thank you for pointing out more tags we should document

In c088ea9698f5164542761078b501a6b426262bf7 I added partial doc for the following tags: 

- `<classloader>`
- `<psr4>`
- `<requires>`
- `<ext>`

I was able to describe `<requires>` and `<ext>`, but unfortunately I don't understand how `<classloader>` and `<psr4>` work (even after reading the above convo between you and @seamuslee001). So I'm not sure how to describe what those tags do. I'm going to merge this PR now, since I think we've had enough eyes on it, but I'd love it if one of you could go fill in the missing info for those two tags. Right now it just says *Not yet documented*.

seanmadsen:
@totten and @seamuslee001 here are the places where you can expand on *"Not yet documented"*.

* https://docs.civicrm.org/dev/en/latest/extensions/info-xml/#classloader
* https://docs.civicrm.org/dev/en/latest/extensions/info-xml/#psr4

totten:
Thanks @seanmadsen !

PR #265, opened 2017-08-11
Migrate "Add-on Formats" wiki page

seanmadsen:
From: https://wiki.civicrm.org/confluence/display/CRMDOC/Add-on+Formats

Reviews welcome. I'll merge this on 2017-08-16 if there are no comments. 

I had to transpose the table to get it to fit on the page. Works ok. Still a bit awkward. Also open to other suggestions for where to put this content in the Dev Guide. I feel like it doesn't fit perfectly into the Extensions chapter, but I couldn't find a place where I thought it would fit better.


totten:
That table does have a ton of information. Transposing seems like a pretty reasonable adaptation for the medium.

At risk of nitpicking over words, I'd focus more on words like "Packages", "Packaging", "Formats" or "Types" rather than "Strategies".  The basic point of the matrix is "for all these Civi APIs, compare how well they're supported in different packaging formats". "Strategy" sounds like a bigger thing to me.

seanmadsen:
@totten I like your changes. You changed the page title in the page. I took this change a step further by also making the same change in the file name and the navigation menu.

seanmadsen:
Merging now, since it seems like Tim has had at least a brief look at this

PR #266, opened 2017-08-12
Migrate Region Reference from wiki

seamuslee001:


seanmadsen:
@seamuslee001

I added some small fixes

Did you notice that this wiki page has a [sub page "HTML Header Region"](https://wiki.civicrm.org/confluence/display/CRMDOC/HTML+Header+Region)? What do you think about putting that page inside this one?

seamuslee001:
Yeh i guess so I'm not sure how much we want to pull out of that page

seanmadsen:
Ahh cool. Yeah if the page is junk then that's fine. I haven't looked at it closely. If there's no useful content in it, then I think we should redirect it to this page in order to kill it.

seamuslee001:
@seanmadsen ported it anyway

totten:
👍 for migrating HTML Region docs.

On a substantive level, it describes some rough stuff. I'm not excited about where the implementation ended-up,  but it is what it is. IMHO understanding it without docs would be an order of magnitude harder. We need docs like that precisely because it's rough stuff.

OTOH, 90% of developers don't really need to understand this -- `CRM_Core_Resources` is the main API. A somewhat better (but hypothetical) home for this might be a reference or guide for how to implement a CMS integration. ("If you're *integrating with a new CMS* or trying to *change the design of the CMS-integration layer*, then here are ten topics to handle/understand. One of them is the HTML header region.")

PR #268, opened 2017-08-12
Add in some documentation on classloader and psr4

seamuslee001:
@totten can you check and make sure this looks correct? @seanmadsen i think this is what is meant it relates to https://github.com/civicrm/civicrm-core/blob/ad16dd2c5345f34330cd6731dd01f7d74218d7b8/CRM/Extension/ClassLoader.php#L117 and https://github.com/civicrm/civicrm-core/blob/648631cd94799e87fe2347487d465b1a7256aa57/CRM/Extension/Info.php#L137

### DIFF
--- a/docs/extensions/info-xml.md
+++ b/docs/extensions/info-xml.md
@@ -81,14 +81,14 @@ Elements acceptable within `<civix>`
 ### `<classloader>` {:#classloader}
 
 * Containing element: [`<extension>`](#extension)
-* Description: This specifies what classes should be added to the Composer ClassLoader
+* Description: Specifies the namespace and file-path of classes to add to the classloader.
 * Contains: elements
 
 Elements acceptable within `<classloader>`
 
 | Element | Acceptable instances | 
 | -- | -- |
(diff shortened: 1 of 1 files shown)


seamuslee001:
Also see http://www.php-fig.org/psr/psr-4/

docs/extensions/info-xml.md
```diff
@@ -250,15 +250,15 @@ Elements acceptable within `<maintainer>`
 ### `<psr4>` {:#psr4}
 
 * Containing element: [`<classloader>`](#classloader)
-* Description: *not yet documented*
+* Description: Specifies the Name space and patho of classes to add the Composer Classloader
```
totten: Typos. Inconsistent caps. We extend composer's class-loader, but that's an implementation detail -- all that matters here is that it's PSR-4 compliant.

Maybe: "Specifies the namespace and file-path of classes to add the classloader."

seamuslee001:
@totten thanks Tim, have made some changes

PR #277, opened 2017-08-14
Migrate PseudoConstant Reference Document

seamuslee001:


docs/framework/pseudoconstant.md
```diff
@@ -0,0 +1,86 @@
...
+| `api.getoptions` | 4.3 | In 4.3 - retrieve option list for a field - works for many but not all fields. In 4.4 - Api wrapper around `CRM_*_BAO_*::buildOptions ` - works for all fields and passes additional params to buildOptions. | `civicrm_api3('contact', 'getoptions', array('field' => 'gender_id'));` Note: accepts additional params in 4.4. See below. |
+
+#### Context
+
+The format of an option list needs to be slightly different depending on the context you're working in. For example if you have machine names as input and need to store the values in the database, you need the list to contain ids and names instead of ids and translated labels. The BAO `buildOptions` method (and the `api.getoptions` wrapper) accept a 'context' string which controls formatting. Quoting from the `CRM_Core_BAO::buildOptionsContext` code:
+
+ - `'get` -  "All options are returned, even if they are disabled.  Labels are translated.",
+ - `create` - "Options are filtered appropriately for the object being created/updated. Labels are translated.",
```
mattwire: So this returns labels?  Also not clear how the filtering works - example?

docs/framework/pseudoconstant.md
```diff
@@ -0,0 +1,86 @@
...
+| `CRM_*_BAO_*::buildOptions` | 4.4 | Retrieve a list of options for a field | `CRM_Contact_BAO_Contact::buildOptions('gender_id');` Note: accepts additional params related to context. See below. |
+| `api.getoptions` | 4.3 | In 4.3 - retrieve option list for a field - works for many but not all fields. In 4.4 - Api wrapper around `CRM_*_BAO_*::buildOptions ` - works for all fields and passes additional params to buildOptions. | `civicrm_api3('contact', 'getoptions', array('field' => 'gender_id'));` Note: accepts additional params in 4.4. See below. |
+
+#### Context
+
+The format of an option list needs to be slightly different depending on the context you're working in. For example if you have machine names as input and need to store the values in the database, you need the list to contain ids and names instead of ids and translated labels. The BAO `buildOptions` method (and the `api.getoptions` wrapper) accept a 'context' string which controls formatting. Quoting from the `CRM_Core_BAO::buildOptionsContext` code:
+
+ - `'get` -  "All options are returned, even if they are disabled.  Labels are translated.",
```
mattwire: Returns all options.  As labels?

docs/framework/pseudoconstant.md
```diff
@@ -0,0 +1,86 @@
...
+ - `create` - "Options are filtered appropriately for the object being created/updated. Labels are translated.",
+ - `search` - "Searchable options are returned. Labels are > translated.",
```
mattwire: Returns labels for enabled options?

docs/framework/pseudoconstant.md
```diff
@@ -0,0 +1,86 @@
...
+ - `validate` -  "All options are returned, even if they are disabled.  Machine names are used in place of labels."
```
mattwire: This one is clearer.

docs/framework/pseudoconstant.md
```diff
@@ -0,0 +1,86 @@
...
+ - `abbreviate` - "enabled options are returned; labels are replaced with abbreviations."
```
mattwire: How are the abbreviations generated? Are they translated

mattwire: In general, I struggle to understand these descriptions.  What I need to know as a developer is exactly what it returns.  A comment and table like this would be helpful:

"buildOptions returns an array of options, keys are option IDs, values are textual:"
`
Context | Returned Options | Text

get | All (including disabled) | Translated labels

create | Some kind of filter | Translated labels

search | Options marked searchable | Translated labels

validate | All (including disabled) | Machine names

abbreviate | Only enabled options | Labels with abbreviations.
`

Aren't we missing a context to return only enabled options as labels, unless abbreviate does that

docs/framework/pseudoconstant.md
seamuslee001: @mattwire good points i think looking at this https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/PseudoConstant.php#L213 suggests to me it would return only enabled options

docs/framework/pseudoconstant.md
seamuslee001: Yep all good points so i think what needs to make difference is the diff between create and search contexts. It seems that both exclude disabled but what does searchable mean. 

I note that according to this https://github.com/civicrm/civicrm-core/blob/f6f65fc0eb2d4440a2662debef1b458ccf836da2/CRM/Core/DAO.php#L2269 there is another context to be documented

seamuslee001:
@mattwire Thanks for the feedback, I hope to have improved the documentation now, Agree it has been confusing, hopefully this gives us a chance to clear up some confusion

seanmadsen:
@mattwire Thanks for your comments here. I've merged this even though I think your comments haven't been fully addressed. I agree that it'd be nice to have more docs which better explain PseudoConstants. For now though, I want to prioritize moving forward with importing wiki content as-is. Adding quick adjustments during the import is great, but I don't want to intermingle importing with larger improvements &mdash; just for the sake of finishing the import.

Issue #81, opened 2017-02-01
Import QuickForm reference


michaelmcandrew:
I know its terribly old fashioned and all but we should have the [QuickForm reference](https://wiki.civicrm.org/confluence/display/CRMDOC/QuickForm+Reference) in the developer guide.

Any idea where this would go into the current structure? I've looked on https://docs.google.com/spreadsheets/d/1ZPq34FuYLBhEzXM2iEGYVZUFLbVZ2CpA3BknLpeHqCQ/edit#gid=0 but I didn't see a proposed high level layout for the developer docs - did I miss something?

Issue #82, opened 2017-02-01
Should quickform reference mention buildQuickForm, not buildForm


michaelmcandrew:
I know this doc issue is out of sync with most of the others being posted but i was looking at https://wiki.civicrm.org/confluence/display/CRMDOC/QuickForm+Reference#QuickFormReference-TheBasics and it mentioned buildForm as the function to use to build the form. I think buildQuickForm is potentially the better choice (civix starts you off with that as well).

I did ask this in the dev channel https://chat.civicrm.org/civicrm/pl/uhqcr5if9389f89ayzh36i13ua but this seems like a safer place for it.

michaelmcandrew:
see also https://github.com/civicrm/civicrm-dev-docs/issues/81

PR #279, opened 2017-08-14
Migrate Quickform Reference Documents

seamuslee001:


seanmadsen:
This will close #81

seamuslee001:
This should also solve #82

PR #284, opened 2017-08-15
Minor style changes

seamuslee001:


docs/framework/quickform/index.md
```diff
@@ -68,8 +68,8 @@ class CRM_Example_Form extends CRM_Core_Form {
 }
 ```
 
-!!!note
-In CiviCRM Core, sometimes in sub classes e.g. in the Custom Search Sub classes, the function `buildForm` is used instead of `buildQuickForm, However there will be a call in the parent function for `buildQuickFrom` which then includes the subClasses buildForm. This is not standard for new form classes.`
+!!!note builQuickForm v buildForm
```
seanmadsen: @seamuslee001 Can you fix this admonition to use [this syntax](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#admonition-syntax)? Specifically:

* Add one space after the exclamation marks
* Add double-quotes around the custom title
* Indent the content by four spaces

MkDocs is picky about these admonitions. It won't display unless you follow those rules.

I've been fixing this syntax from you in other PRs, which is usually faster than writing a comment about it. I bring it up now since you're doing enough docs work where it's worth it for me to start pointing more things out. :slightly_smiling_face:

seamuslee001:
@seanmadsen sorted now

PR #285, opened 2017-08-16
Migrate "Database standards" wiki page

seanmadsen:
Migrate this page: https://wiki.civicrm.org/confluence/display/CRMDOC/Database+layer

Most of the content on this wiki page meets the level of quality I would expect out of a note scribbled onto a napkin after an informal discussion at the pub. It's not documentation.

Nonetheless, I [discussed](https://chat.civicrm.org/civicrm/pl/hpzkpu7ck7g9jnqh1rrqeqh1gw) this page with @eileenmcnaughton and she said: 

> I think the DB layer one is probably mostly right down to the compliance section which is dated & can go

Still, I ended up deleting *everything after the word "compliance"* &mdash; which included several other sections on the page. That left... not much. But here we are. One more page to cross of the list! 

Any feedback welcome on this one! 

Will merge on 2017-08-18 if no reviews.


eileenmcnaughton:
@seanmadsen I think this is fine to merge as is. The points in it are correct in terms of specific consistenency we would like to have in our code (but often do not).

seanmadsen:
Thanks @eileenmcnaughton. Merging.

PR #288, opened 2017-08-16
Fix typos

toraritte:
Thanks for this awesome tool! I work for a blindness non-profit and we are in the middle of creating an client management system and this would perfectly fit for our resource dev team. I have long been mulling over the DB design, getting stuck with how we could attach notes with different entities and it is good to see that the idea we've come up with works in production:) (entity_id, entity_table etc.)

Have a great day!
Attila


seanmadsen:
Thanks for these fixes @toraritte and great to hear about the work you are doing! :slightly_smiling_face:

PR #289, opened 2017-08-16
Fix links

toraritte:
Thanks again and have a great day!

### DIFF
--- a/docs/framework/database/schema-design.md
+++ b/docs/framework/database/schema-design.md
@@ -13,10 +13,10 @@ Another method of viewing CiviCRM's Database model is to use one of various MySQ
 1. Look at the list of tables in the database. The table name generally gives you a good idea of the type of data it stores.
 2. Pick one table and look closer. Let's look at `civicrm_country` (a relatively simple example) by opening the "Table inspector" for that table.
     - We can look at the list of columns in the table.
-    ![MySQL Workbench CiviCRM Country Columns](/img/mysql_workbench_civicrm_country_tables.png)
+    ![MySQL Workbench CiviCRM Country Columns](/docs/img/mysql_workbench_civicrm_country_tables.png)
     We can see that columns have comments which explain (to some extent) the meaning of the data stored in the columns. From the Column Names and the comments we can start to get some idea of the relationship between tables. For example we can see that there is a column `address_format_id` which indicates to us that the table `civicrm_country` has a relationship to the table `civicrm_address_format`
     - We can also look at the Foreign Keys that are relevant to this table
-  ![MySQL Workbench CiviCRM Country Foreign Keys](/img/mysql_workbench_civicrm_country_foreign_keys.png)
+  ![MySQL Workbench CiviCRM Country Foreign Keys](/docs/img/mysql_workbench_civicrm_country_foreign_keys.png)
   Here we can get more of a comprehensive picture of what tables `civicrm_country` relates to. We can tell that because we see not only what tables are referenced by `civicrm_country` but also the tables that reference `civicrm_country`. 
 
 ## General Characteristics of CiviCRM Tables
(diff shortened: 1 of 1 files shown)


docs/framework/database/schema-design.md
```diff
@@ -13,10 +13,10 @@ Another method of viewing CiviCRM's Database model is to use one of various MySQ
 1. Look at the list of tables in the database. The table name generally gives you a good idea of the type of data it stores.
 2. Pick one table and look closer. Let's look at `civicrm_country` (a relatively simple example) by opening the "Table inspector" for that table.
     - We can look at the list of columns in the table.
-    ![MySQL Workbench CiviCRM Country Columns](/img/mysql_workbench_civicrm_country_tables.png)
+    ![MySQL Workbench CiviCRM Country Columns](/docs/img/mysql_workbench_civicrm_country_tables.png)
```
seanmadsen: @toraritte I'm a little curious about this change. When you go to https://docs.civicrm.org/dev/en/latest/framework/database/schema-design/#mysql-workbench-workflow do you find that the "Country Columns" screenshot is not loading? Because I'm observing it loading (with the current path that begins with `/img`). I'm worried that adding `/docs` to the beginning of the path with cause the screenshot to no longer load. But if you're observing problems with the image loading, then I'd like to try and reproduce that on my end.

docs/framework/database/schema-design.md
```diff
@@ -26,6 +26,6 @@ When we look at CiviCRM Tables there are a few patterns that generally hold true
 - Every table has a id column which is Auto Increment and therefore unique key for that table
 - Columns which reference other tables generally speaking will be named in the format of `other table name` + `_id`. For example in `civicrm_country` there is `address_format_id` which indicates that is references `civicrm_address_format.id`
 - Many-to-many relationships use "join tables" as intermediary tables. For example, a contact can have many activity records, and an activity can have many contact records. So the table `civicrm_activity_contact` is used as the glue because it has foreign keys to both.
-- In some places CiviCRM defines schema using a construct called [PseudoConstants](/framework/database/schema-definition.md#table-field-pseudoconstant) which produces some slightly more complex logic
+- In some places CiviCRM defines schema using a construct called [PseudoConstants](/docs/framework/database/schema-definition.md#table--field--pseudoconstant-table-field-pseudoconstant) which produces some slightly more complex logic
```
seanmadsen: Similarly, this link *is* working for me as-is. Is it not working for you?

Note this point [here](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#internal-url-standards): 

> Use absolute URLs which begin with a forward slash. (The root directory is the `docs` folder.)

Perhaps I should highlight that point more prominently? I realize that perhaps it's confusing.

docs/framework/database/schema-design.md
toraritte: Your links does work properly. I was reading the docs from github (https://github.com/civicrm/civicrm-dev-docs/blob/master/docs/framework/database/schema-design.md) and the links would give me 404 but now I just realized that this repo generates the docs for docs.civicrm.org... 

You can discard this pull request, I'm sorry for taking your time.

seanmadsen: Ahhh. Interesting. Ok that makes a lot of sense! Thanks for letting me know about this. I'm going to think on this and see if I can identify a way to prevent confusion like this in the future.

toraritte: I don't think that there is anything wrong on your end because this is the repo for your online documentation and it is not the supposed to be main source. As soon as I heard of this project, I immediately looked up the CiviCRM repos, found the dev docs, then I got excited and over-zealous whereas most of your visitors would probably find these topics the "proper" way:)

Thanks again!

PR #291, opened 2017-08-17
Change to actual function that exists

mattwire:


tschuettler:
It seems to be used more prominently used as ```CRM_Core_Session::getLoggedInContactID()``` ?
https://github.com/civicrm/civicrm-core/pull/9227/files

PR #295, opened 2017-08-19
Migrate CiviMail Reference Document

seamuslee001:
@totten  Tim i have requested a review from you on this one as i have significantly re-written the token section compared to the current document to be more on how does it build the tokens etc and i know that you have had work done on this . 

There is probably more work to expand the events but this is a good first stab at it. 

I'm not sure if we need to in here go into transactions as how CiviMail runs things but i feel like that is a bit of a different document but happy for comments


docs/framework/civimail.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Tokens
+
+In 4.7+ there was major changes to the Scheduled Reminders facility which also included changes to CiviMail in so far as how tokens are generated. There is a move to use more of the `Civi\Token\TokenProcessor` sub system as this is more robust. However there have been compatibility layers built in to use the older `CRM_Utils_Token` processors. Developers should aim to work off the `Civi\Token\TokenProcessor` where possible. However there are still some systems that haven't been refactored. Some of the key functions in the older systems are. 
+
+- `CRM_Utils_Token::getTokens` - Retrieves an array of tokens contained in the given string e.g. HTML of an email
+- `CRM_Utils_Token::getRequiredTokens` - What are the minimum required tokens for CiviMail
+- `CRM_Utils_Token::requiredTokens` - Check that the required tokens are there
+- `CRM_Utils_Token::&replace<type>Tokens` - Replaces x type of Tokens where x is User, Contact, Action, Resubscribe etc
+- `CRM_Utils_Token::get<type>TokenReplcaement` - Format and escape for use in Smarty the found content for Tokens for x type. This is usually called within `CRM_Utils_Token::&replace<type>Tokens`
```
mattwire: TokenReplcaement->TokenReplacement

docs/framework/civimail.md
```diff
@@ -0,0 +1,155 @@
...
+
+Job `status` can take one of 5 states.
+
+1.  `Scheduled`: All jobs start in this state, and stay there until the `start_date` has passed.
+2.  `Running`: Jobs are marked `Running` only after the entire recipient queue (`Mailing_Event_Queue`) has been constructed. If the queuing process is interrupted, the job will remain in `Scheduled` and the queue will be reconstructed the next time the mailer is run.
+3.  `Complete`: Jobs are marked complete after the mailer has attempted to send the message to every recipient in the queue (ie. every queue event has a corresponding bounce or delivery event). Note that slow bounce events may continue well after the job has been marked `Complete`.
+4.  `Paused`: A job can only be marked paused by the admin interface. The mailer will not act on paused jobs.
+5.  `Canceled`: Like paused, but cannot be placed back in the `Running` state.
+
+## Events
```
mattwire: Not sure "events" is quite the right name here, but not sure what else to suggest.  May also be useful to note somewhere here that an activity is created against each contact with a link to the mailing

seamuslee001:
@mattwire thanks for the review, wrt the Events section, yeh i agree but i'm not sure what to do i have re-named it now but it also feels strange as all those things are run out of `CRM/Mailing/Event/xxx` in the codebase

seanmadsen:
@seamuslee001 I fixed this myself so that we could get this merged but take a look at the fix I made in 6f2079a77be9c2a1830690d0923623beb708e42f. The markdown rendering engine that we use requires that we put 4 spaces before sub-lists. Otherwise they don't show up as sub-lists. Mentioning now so that you're aware of it going forward.

Issue #299, opened 2017-08-20
Find/replace hyperlinks to wiki with internal links


seanmadsen:
After the wiki import we should do the following:

1. Look for links which point to the wiki
1. If the link is also present in the wiki redirects file, then replace the wiki link with the correct local link

I think we could probably do this with a script.

seamuslee001:
In regards to 1

These are all the locations that use  wiki links excluding that of what i have submitted in a PR just now

```
docs/api/usage.md:[Bootstrap Reference]: https://wiki.civicrm.org/confluence/display/CRMDOC/Bootstrap+Reference
docs/api/usage.md:[API Security](https://wiki.civicrm.org/confluence/display/CRMDOC/API+Security).
docs/api/usage.md:[AJAX Interface]: https://wiki.civicrm.org/confluence/display/CRMDOC/AJAX+Interface
docs/api/usage.md:[API Security](https://wiki.civicrm.org/confluence/display/CRMDOC/API+Security)
docs/api/usage.md:[Smarty API interface](https://wiki.civicrm.org/confluence/display/CRMDOC/Smarty+API+interface).
docs/api/changes.md:See [EntityRef Fields documentation](https://wiki.civicrm.org/confluence/display/CRMDOC/EntityRef+Fields) for more info.
docs/api/changes.md:See [Pseudoconstant (option list) Reference](https://wiki.civicrm.org/confluence/display/CRMDOC43/Pseudoconstant+%28option+list%29+Reference)
docs/api/changes.md:See [Pseudoconstant (option list) Reference](https://wiki.civicrm.org/confluence/display/CRMDOC/Pseudoconstant+%28option+list%29+Reference)
docs/api/changes.md:When creating a new Case record, the "create" API previously accepted `case_type` using the displayable title of the case-type (e.g. "Housing Support"). Beginning in v4.5, the `case_type` should specify the internal machine-name of the case-type (e.g. `housing_support`). Submitting the displayable title will still work - but this is deprecated. For more discussion about changes to handling of case-type names, see [Check Case Type Name Consistency](https://wiki.civicrm.org/confluence/display/CRMDOC/checkCaseTypeNameConsistency).
[...]

PR #305, opened 2017-08-22
Fix Wiki URLS to point to correct docs url

seamuslee001:


seamuslee001:
Merging following discussion with @seanmadsen he was ok with this merge as is but noting this is only part of a bigger effort #299

PR #312, opened 2017-08-27
Add in some documentation on PHPIDS

### LINKED ISSUE #281: Incorporate Documentation on IDS in Security Docs
@totten @seanmadsen i was thinking that for a 2nd pass after the Security Docs PR is merged, we should look at including the IDS system in someway as it is a security feature unless its already covered and i didn't see it. I think it would be useful to have more documentation on it.

seamuslee001:
@totten @seanmadsen can you take a look at this, I think this is in the right location for it as it is about inputs

This should resolve #281


seanmadsen:
@seamuslee001 Thanks for this. Excellent addition! 

I did some minor restructuring to clarify that PHPIDS actually applies to *all* inputs (even things like `GET` params which aren't necessarily saved to the DB).

PR #317, opened 2017-08-30
hook_civicrm_angularModule - Add basic doc for `settings`

totten:
This fills in a little info about an additional field named `settings`. 
It's not exactly as deep as "War & Peace", but it's about the same level as
similar docs.


eileenmcnaughton:
Yay. @GinkgoFJG fyi

seamuslee001:
This looks right and i think it makes sense. @seanmadsen

PR #325, opened 2017-09-03
Document civi-test-run script

seamuslee001:
The aim of this is to document the new script here https://github.com/civicrm/civicrm-buildkit/pull/346/files Tim are you able to check this please

### DIFF
--- a/docs/testing/javascript.md
+++ b/docs/testing/javascript.md
@@ -10,8 +10,14 @@ conventions using [karma] and [jasmine].
 ## Running Javascript Tests
 
 ```bash
-cd /path/to/civicrm
-npm test
+$ cd /path/to/civicrm
+$ npm test
+```
+
+You can also run the karama tests as they would be run by Jenkins using the `civi-test-run` command
+
+```bash
+$ civi-test-run "<buildname>" "<civicrm version>" "<junitdir>" "karma"
 ```
 
 [karma]: https://karma-runner.github.io/1.0/index.html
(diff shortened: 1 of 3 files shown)


seanmadsen:
Docs-wise, this looks good to me regarding, spelling, formatting, placement. I haven't reviewed it for conceptual accuracy, since I don't have time right now to dive into the buildkit PR. 

I feel ready to merge this when the related PR is merged.

totten:
Notes:
 * The signature for `civi-test-run` changed  --the `<version>` is now autodetected.
 * We've merged `civi-test-run`, but I'd still treat the signature as [experimental / subject-to-change](https://github.com/civicrm/civicrm-buildkit/pull/346#issuecomment-328695083).
 * I suppose the rationale for documenting two ways side-by-side is that there are a couple audiences looking at this:
    * If you're an IDE user, then it's better to see the *vanilla/upstream* test-runner (eg `phpunit` or `karma run`) because that'll help you configure the IDE integration. 
    *  If you're setting up CI, then it's better to see the *comprehensive* test-runner (eg `civi-test-run`) because it's easier to put all the suites into your CI job.

seanmadsen:
@totten I see the related PR has been merged but I don't really have enough time to dig into it deeply enough to fully understand your last comment. 

I *think* you're saying that we need to change

```bash
$ civi-test-run "<buildname>" "<civicrm version>" "<junitdir>" "karma"
 ```

to

```bash
$ civi-test-run "<buildname>" "<junitdir>" "karma"
 ```

before merging this PR. Is that right?

**Comment by totten** [comment]
The immediate change to match the current code would be as you describe:

```
$ civi-test-run "<buildname>" "<junitdir>" "karma"
```

But it goes a little further than that -- the new helper is still *experimental*, and the signature is likely to change again, eg:

```
$ civi-test-run -b "<buildname>" -j "<junitdir>" "karma"
```

In this case, we merged (despite being experimental) because we need that in order to give the functionality a trial run on Jenkins.

seanmadsen:
To address above comments from @totten (as well as some other thoughts I had about this PR as I dug deeper), I create #367 to replace this PR. 

I am proposing we close this PR in favor of #367.

PR #330, opened 2017-09-07
Document how to load new Settings metadata

xurizaemon:


totten:
Good idea.

Relatedly, there's a blurb about the various ways to clear the cache at https://docs.civicrm.org/dev/en/latest/tools/debugging/#clearing-the-cache . Might be worth linking.

Although I'm not sure -- can you do internal links with anchors? like 

```
[clearing the cache](tools/debugging/#clearing-the-cache)
```

seanmadsen:
> can you do internal links with anchors?

Yep, although it needs to follow [this syntax](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#internal-url-standards)

PR #331, opened 2017-09-07
Add example of PSR-4 autoloading within extensions

seanmadsen:


seamuslee001:
This looks good to me and i think it make sense in the context so merging

PR #334, opened 2017-09-08
Update hacking.md

dvhirst:
Add link to material that describes "talking with the CiviCRM developer community".


seanmadsen:
I like this, but I think the link needs to be fixed to use `basics` as the directory instead of `the-civicrm-community`. Can you update that?

Note: if you need help figuring out how to update an existing branch of yours ping me in Mattermost.

dvhirst:
Done, not elegantly, but it worked.  I deleted the pull request and created a new one with the more appropriate link.


Don



-----Original Message-----

PR #351, opened 2017-09-09
Update pr-review.md

dvhirst:
Reading for content; editing for consistency. 
Note:  Check for Drupal dependencies in the DB upgrade case.

### DIFF
--- a/docs/core/pr-review.md
+++ b/docs/core/pr-review.md
@@ -27,11 +27,9 @@ If you're a beginning developer looking for easy PRs to review, you might have g
 
 Add a comment to the PR like "Reviewing this now" to let others know that you intend to submit a review.
 
-
 ## Read about the issue
 
 Every PR *should* have an issue ID for [JIRA](https://issues.civicrm.org) linked from the PR's page on GitHub. Read the original issue and understand how to reproduce the problem and what the solution looks like as well.
-
 
 ## Read the code changes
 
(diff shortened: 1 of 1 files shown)


docs/core/pr-review.md
```diff
@@ -87,8 +83,8 @@ An easy way to do this is:
 
 1. Install a buildkit site for the latest publicly available release of CiviCRM (*not* `master`). Pass the `--civi-ver` option to civibuild for this.
 1. Update the `civicrm` directory files so that the codebase has the changes in the PR (perhaps by using [Hub](https://hub.github.com/) as described above).
-1. From the `civicrm` directory, run `./bin/setup.sh -Dg` to update the generated-code
-1. Run `drush civicrm-upgrade-db` to perform database upgrades
+1. From the `civicrm` directory, run `./bin/setup.sh -Dg` to update the generated-code.
+1. Run `drush civicrm-upgrade-db` to perform database upgrades. *(TODO:  Check for Drupal dependencies.  If Wordpress or Joomla! is the target CMS, how should database upgrades be accomplished?)*
```
seanmadsen: @dvhirst it seems like your "TODO" comment here is intended to be a to-do item for docs editors (not readers). Am I correct? If so, I'd rather we leave this content out of the guide and instead open a ticket within this repo to track this improvement. This way readers will have a more polished experience and editors will be more likely to spot (and fix) the issue. Is that okay? If you're cool with this approach, then I think it'll merge this PR after you remove this "TODO" bit.

dvhirst: I'm OK with your approach, and will remove the TODO item.  You'd asked me to highlight Drupal dependencies as I found them, this was the initial one I spotted.  In future, I'll mark them with a new issue ticket in the appropriate docs folder.

dvhirst:
Editorial TODO removed...

seanmadsen:
Awesome. Thanks Don.

PR #363, opened 2017-09-11
api/changes - Clean up formatting under "4.7.0: Global $civicrm_setting"

totten:
Each commit includes a more detailed explanation.


seamuslee001:
To me this looks fine and i think better explanation, @totten have you thought of documenting the `getMandatory` method as well? @seanmadsen does this look ok to you?

totten:
Re: `getMandatory()` -- it would make sense to mention it in https://docs.civicrm.org/dev/en/master/framework/setting/ but it doesn't need to be tied into this cleanup of the API changelog. Proposed a separate issue #364.

seanmadsen:
* Changes look good to me
* I agree that #364 is a good way to isolate that improvement and move forward with this one separately.

Merging

PR #367, opened 2017-09-12
Document civi-test-run script (take 2)

seanmadsen:
This builds on top of #325 and is intended to replace it. 

In #325 Seamus added docs for a new script, `civi-test-run`. 

This PR is different from #325 in the following ways: 

* Use a separate page for `civi-test-run` in the "Tools" chapter
* Point readers to this page so that we can avoid duplicating efforts to describe `civi-test-run`
* Also mention `civi-test-run` in the *list* of all tools
* Don't describe the script's usage. Instead instruct readers to run the script to observe its usage via the command line. (This is to address [comments](https://github.com/civicrm/civicrm-dev-docs/pull/325#issuecomment-328699764) from Tim about how the signature is likely to change in the future)

### DIFF
--- a/docs/testing/continuous-integration.md
+++ b/docs/testing/continuous-integration.md
@@ -35,5 +35,9 @@ codebase.
  
 Code-style is not checked on this build, but all upgrade and web tests are run.
 
+## Running the test suite locally
+
+You can use [civi-test-run](/tools/civi-test-run.md) locally to run a full standard CiviCRM Test suite.
+
 [jenkins-test-results]: https://test.civicrm.org/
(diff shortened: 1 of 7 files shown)


seanmadsen:
Would like thoughts from @seamuslee001 @totten before merging

totten:
+1 for moving to its own page.

I've tweaked to mention the `upgrade` and `karma` suites.

seanmadsen:
I like the tweaks, Tim.

I'll merge this on 2017-09-14 if no further comments. 

@seamuslee001 feel free to merge if you approve &mdash; or push more commits or suggest changes.

seamuslee001:
Yep this looks good to me merging

PR #368, opened 2017-09-12
Add callback example

mattwire:
This caused me a few hours of pain today.  Hopefully this will save someone else that time!


seanmadsen:
Thanks @mattwire 

I made a couple tweaks. 

Note that the markdown parser is pretty strict about the [syntax for admonitions](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#admonition-syntax).

PR #370, opened 2017-09-14
Restore alternative options to icons list

agh1:
In the migration, the third column of the icon documentation from the wiki was lost.  This part is valuable because it helps developers find the right icon to use when they are thinking of something that seems close.

Precision in choosing icons is what helps us avoid having the same checkbox icon used for everything from "yes, that's fine" to "save my work" to "please make big database changes".


seanmadsen:
Thanks @agh1 !

Whoops sorry I dropped that column. My bad!

This is a big improvement. Great idea to link to the Font Awesome page for each icon. 

Side note: I spent a small amount of time trying to actually display the icons within the docs, but gave up. I think it's still possible, and could be neat. For another day...

PR #371, opened 2017-09-14
Corrected git guidance.

GinkgoFJG:


docs/extensions/publish.md
```diff
@@ -43,7 +43,7 @@ CiviCRM's publishing process automates a number of tasks related to maintaining
 -   Create a git "tag" which matches the version in the manifest and push it, e.g.:
 
         git tag -a v1.2.0
-        git push origin v1.2.0
+        git push --tags origin v1.2.0
```
totten: Interesting, I've never used that option. The manual describes it as:

```
       --tags
           All refs under refs/tags are pushed, in addition to refspecs explicitly listed on the command line.
```

which suggests to me that you can either

 * Push tags one at a time (`git push origin v1.2.0`)
 * Push tags en masse (`git push --tags`)

GinkgoFJG:
Doh, I think you're right about that. I think we can withdraw this PR.

PR #372, opened 2017-09-17
CRM-21127 - Document fixed return properties in api.Event.get

colemanw:
Adds documentation for https://github.com/civicrm/civicrm-core/pull/10984


seanmadsen:
Thanks Coleman

PR #373, opened 2017-09-18
Migrate CiviTest Examples for two hooks that still reference them

seamuslee001:


seamuslee001:
Thanks @seanmadsen just didn't want to do the whole self merge thing

PR #375, opened 2017-09-19
Update documentation to take into account different apache conf folde…

### LINKED ISSUE #374: Buildkit install docs assume `/etc/apache2/conf.d`
My Apache knowledge is not up to date, but I believe there are some distros which don't use this directory. I've seen `/etc/httpd` and I've also seen `/etc/apache2/conf-{available,enabled}`.

In [this ~dev-newcomers](https://chat.civicrm.org/civicrm/pl/oz8k5tip3f8z5qskucbhgjpane) chat, @hassan3152 followed instructions accurately but the results were unsuccessful.

The instructions [as per current](https://github.com/civicrm/civicrm-dev-docs/blob/7034a0df604e77cc9342b5bd164391ae8d7491f8/docs/tools/buildkit.md) are really good, but not helpful for someone with an Apache layout not matching the docs.

I'm not sure how best to rephrase this.

We probably can't try to handle *all* possibilities (`/usr/local/etc/{httpd,apache2}`, `/opt/local/apache2/conf/`, ...). But we can tell people to find the right directory first?

seamuslee001:
…r locations

I think this should fix #374 @xurizaemon are you able to look over this and see if it makes sense

### DIFF
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -126,19 +126,23 @@ Buildkit provides a tool called `amp` which [civibuild](/tools/civibuild.md) use
     
         * For Apache 2.2: 
         
-            Create a new file `/etc/apache2/conf.d/buildkit.conf` with the following contents:
+            Create a new file `buildkit.conf` with the following contents in either `/etc/apache2/conf.d/` or `/etc/httpd/conf.d` depending on your server configuration.
         
             ```
             Include <amp-installation>/apache.d/*conf
             ```
+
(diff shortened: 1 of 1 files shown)


seanmadsen:
This is great. I added a commit that build on top of work from @seamuslee001 by re-orgainzing the steps a little bit to remove duplication and make things a bit more explicit.

seanmadsen:
@xurizaemon can you give this a glance to see if it addresses your intent with #374 ?

PR #390, opened 2017-10-07
toosl/buildkit.md - Don't talk so much about nginx

totten:
As noted earlier in the doc: "We strongly recommend using Apache as your
webserver because support for nginx is limited." So...  it doesn't make much
sense to give broken/incomplete instructions for something we say you
shouldn't do.  :)


seanmadsen:
I think this makes sense. If we change buildkit to better support nginx at some point in the future, we can add more docs about it. Until then, nginx users are a bit more on their own.

PR #392, opened 2017-10-08
Little typo

J0000:


seanmadsen:
Thanks for this PR @J0000 

Are you sure this is a typo though? I don't know off-hand either way, but I can imagine it's possible that the object name would be a CamelCase string. I would feel okay merging this if: 

* You tried with `'Contact'` and it *didn't* work **and**
* You tried with `'contact'` and it *did* work

seamuslee001:
@seanmadsen i think there was some chatter about this in mattermost in i think the town square maybe on this

seamuslee001:
Actually it was dev-newcomers see discussion here https://chat.civicrm.org/civicrm/pl/9wffkekchbyy9pb6t5fh8u6uer

J0000:
I tried with 'Contact' and it didn't work and I tried with 'contact' and it did work on CiviCRM 4.6

seanmadsen:
Ok cool. Merging now based on last comment from @J0000

Issue #397, opened 2017-10-08
Properly spell WordPress in the docs


kcristiano:
There are a number of files that have used 'Wordpress' instead of 'Wordpress' in the documentation.  We should replace 'Wordpress' with 'WordPress' to use correct camel casing according to the WordPress Wordmark and Trademark

seamuslee001:
PR has now been merged closing

PR #398, opened 2017-10-08
Issue # 397 Replace 'Wordpress' with 'WordPress' to use correct camel casing acco…

kcristiano:
…rding to the WordPress Wordmark and Trademark  https://github.com/civicrm/civicrm-dev-docs/issues/397


seamuslee001:
I'm happy with this if @seanmadsen is also. It does seem to be the correct thing here

PR #402, opened 2017-10-08
Docs are a repeat of the script output -- 2 places to keep it up to date???

marsh-circle:
This may generate discussion, and is a suggestion - the `amp configure` command detail here is given in the command output (and so creates two places to keep up-to-date) and  gives a different workflow (so creates confusion), so:
    
I've taken out the repetition, and drawn attention to the script output in the `!!! tips` that could well be overlooked.


seanmadsen:
@marsh-circle (and CC @wmortada )

Tim and I looked at this together this evening and made some changes to `amp config` to provide much better instructions to the user at the step.

In light of this improvement to `amp`, I think it's better to cut these steps out of the docs. Hence, I'll be adding a commit to this PR to cut a little bit *more* out...

seanmadsen:
Actually, closing in favor of #405

PR #405, opened 2017-10-09
Remove instructions for adding amp settings to apache

seanmadsen:
Tim and I added these instructions to amp (and improved them) so now it's better to remove them here.


seanmadsen:
Based on feedback from @marsh-circle and @wmortada installing buildkit, and discussions at the UK sprint between the three of us and @totten, we all thought it best to remove these instructions in favor of putting them directly into amp.

Here is the change to `amp`:

https://github.com/amp-cli/amp/commit/92650fb060929ccb56bc96e7b5f9465f890dec49

Issue #190, opened 2017-06-16
Clarify in Style Guide: shell commands -- prefix with dollar sign or not?


seanmadsen:
When I write documentation that instructs the user to execute a shell command, how should I write it? 

### With the dollar sign

```bash
$ civibuild create dmaster --url http://dmaster
```

### Without the dollar sign

```bash
civibuild create dmaster --url http://dmaster
```

I think it'd be nice if we were consistent on this matter, so I think it'd be a good candidate for putting in the style guide. 

I don't have an opinion either way. Do you?

PalanteJon:
IMO, the main effect this has is to inhibit copy/pasting multiple lines of shell code.

While I can see the argument that it would be a feature, not a bug, to do so, I'm not on board.  While I don't run shell code I don't understand - anymore - there's no doubt I would have never learned all I have unless I was able to get somewhere through the use of copy/paste.  So I would support leaving the dollar sign out.

totten:
Agree with Jon 's easier to copy-paste if you omit the dollar sign.

However, the commands often aren't written in isolation, eg
 * If there are multiple commands, then we might mingle in some comments.
 * If the output is significant to the discussion, then we might show it as part of blob.

If you looked at docs I wrote, you'd probably see inconsistent usage  of `$`. But there maybe a hidden rule which I'm following:
 * If there's output interleaved, then include the `$`.
 * If there's no output interleaved, then omit the `$`. (Unless you're in a page that has lots of other preformatted blurbs, in which case you might adjust to match them.)

For some examples, see my Angular PR:

https://github.com/totten/civicrm-dev-docs/blob/master-angular/docs/framework/angular/quickstart.md#open-the-page

The section "Open the page" is actually pretty carefully considered for readers of multiple perspectives:
 * Providing the visual output gives enough clues that you should be able to figure the URL by yourself. You don't need `cv` at all. (And, at this point in this page, there's no guarantee that you have it installed!) 
 * Providing the command gives a fallback -- a way you can figure out more complicated scenarios.
 * If we just showed one example URL for Drupal, it would be confusing to Joomla/WordPress folks -- they'd incorrectly guess the URL.

totten:
Another variation that shows up: sometimes the PWD is indicated.
 * This is useful since PWD can be significant, and it's a common mistake for novices to mix up their PWD. It pains me to see support requests (there were several in the past) because someone was in the wrong PWD. Showing the PWD helps you error-correct.
[...]

PR #410, opened 2017-10-09
Add in style guidelines on writing shell documentation as per isuse #…

seamuslee001:
…#190 proposal

ping @totten @MegaphoneJon @seanmadsen @nganivet I think Sean's latest proposal on this issue is good so have written documentation based on that but if we need more discussion best go on the issue


seanmadsen:
:+1: I like this. Will wait to merge until we have more approvals from others.

nganivet:
+1, very sound and clearly written.

MegaphoneJon:
This sounds good.

seamuslee001:
@seanmadsen seems like we have consensus here

PR #424, opened 2017-10-12
Typo karama instead of karma

scardinius:


seamuslee001:
Thanks for picking this up merging

PR #425, opened 2017-10-14
Asset builder example

aydun:
Add step-by-step example
@totten - thanks for the train station tutorial!


totten:
I like how "CSS Example" shows the declaration (`hook_civicrm_buildAsset`) and then shows the command to preview it (`cv ev ...`). We should probably update "Usage: Simple asset" to do something similar.

On second thought, a more natural sequencing might be:

 1. Declare the asset (`hook_civicrm_buildAsset`)
 2. Preview the asset (`cv ev ...` )
 3. Use the asset at runtime

PR #426, opened 2017-10-17
change path in Entity creation w/flow

marsh-circle:
I'm not sure if this has only come about because of my ineptitude (i.e. whether an Entity would always have the same name as the extension), but I found that I couldn't find the files referred to because my Entity and my Extension had different names.

Would that be bad practice on my part? (it would be good to know as I'm very new to coding)

Either way, I think this may be more accurate... whether I'm right or not is up to you.

Cheers,
M.

### DIFF
--- a/docs/extensions/civix.md
+++ b/docs/extensions/civix.md
@@ -158,7 +158,7 @@ If you want your extension to store data in the database, then you will need to 
 
         * *Prior to CiviCRM 4.7.12, you would instead run `php ./GenCode.php` from the `xml` folder.*
 
-    1. Copy the DAO file `<civiroot>/CRM/MyEntity/DAO/MyEntity.php` into your extension at `CRM/MyEntity/DAO/MyEntity.php`.
+    1. Copy the DAO file `<civiroot>/CRM/MyExtension/DAO/MyEntity.php` into your extension at `CRM/MyExtension/DAO/MyEntity.php`.
 
     1. Within your extension, create an `sql` directory and create two empty files within it: `auto_install.sql` and `auto_uninstall.sql`.
 


seanmadsen:
Thanks Martin. I don't think this is quite right though. An extension can define many different entities, so it will be pretty common for the name of the extension to be different from the name of an entity it defines. 

More thoughts: 

* For

    > Copy the DAO file `<civiroot>/CRM/MyEntity/DAO/MyEntity.php`

    I think this should definitely be correct because that's talking about a file within CiviCRM *core*. 

* For

    > into your extension at `CRM/MyEntity/DAO/MyEntity.php`

    I think this is correct, but perhaps the docs could make this clearer by saying: 

    > into your extension at `<extension_root>CRM/MyEntity/DAO/MyEntity.php`

    Would that change (adding `<extension_root>`) have helped you?

I'd be curious to hear more about the problem you encountered here.

marsh-circle:
Hi Sean,
When going through the process, the file was definately placed in <civiroot>/CRM/**MyExtension**/DAO/MyEntity.php.

If it were placed in CRM/MyEntity, wouldn't there be a lot more directories in <civiroot>//CRM/* (there would be a directory for every entity?)
Then when the file is moved, it should be moved to <extension_root>**/**CRM/**MyExtension**/DAO/MyEntity.php, as the entity is a part of the extension.
I'm prepared to be wrong, but can't see it at the moment...

seanmadsen:
@seamuslee001 or @michaelmcandrew  would you be willing to give this a look and see what you think?

seamuslee001:
it has been my understanding that it has always been `CRM/MyExtension/DAO/MyEntity.php` for entites in CiviCRM. looking at https://github.com/civicrm/org.civicrm.volunteer you can see they have CRM/Volunteer which matches the CRM/MyExtension structure likewise here https://github.com/eileenmcnaughton/nz.co.fuzion.civixero therefore I believe this is accurate, I think it also might come back to name space stuff i.e. classes are meant to be `CRM_MyExtension_xxx` so there isn't clashes around the place

seanmadsen:
Ok cool. Looks like I was wrong on this one. Sorry Martin! So @seamuslee001 you think this is ready-to-merge then? If so, feel free to merge it yourself.

marsh-circle:
Thanks Sean / Seamus.

Sean - I'm sure there'll be many instances that our roles will be reversed - no apology necessary - although appreciated :)

Cheers :D

PR #430, opened 2017-10-19
standards/php.md - Add a TLDR example

totten:
To see actual examples of the coding style, you need to skim multiple blurbs
and click multiple links, and the eventual page shows very isolated
examples.

This revision adds a brief example so that a new contributor can get a
general sense of the layout immediately.

(Note: It's not possible for a brief example to show every edge-case, and we
shouldn't try -- that's why the commit also emphasizes the link into the
relevant examples from d.o docs.)


seanmadsen:
Great idea!

PR #432, opened 2017-10-23
documentation/style-guide.md - Mention bash portability tip

totten:
If a reader has customized their interactive shell, their system could have
small discrepancies in shell notation.  It shouldn't matter 95% of the time,
and it's not critical to support those folks perfectly.  But it is nice.  So
let's just share the link to heighten awareness.

(Example: I usually do `export VAR=value`...  but in future docs, I'll try
to use `VAR=value; export VAR`.)

This is following-up on some comments from @ginkgomzd who found the
`bash-hackers.org` site and shared it on https://github.com/civicrm/civicrm-buildkit/pull/365


seanmadsen:
Cool :slightly_smiling_face:

PR #434, opened 2017-10-24
mkdocs.yml - Simplify menu headings

totten:
When scanning the navbar, there are several words which (IMHO) aren't adding
much value.  This simplifies and should make it easier to navigate.

## Before

screen shot 2017-10-24 at 3 50 57 pm

## After

screen shot 2017-10-24 at 3 51 11 pm

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -196,7 +196,7 @@ pages:
   - Permissions: security/permissions.md
   - Access Control: security/access.md
   - Reporting Vulnerabilities: security/reporting.md
-- Framework Reference:
+- Framework:
   - AJAX Pages and Forms Reference: framework/ajax.md
   - AngularJS:
     - AngularJS Intro: framework/angular/index.md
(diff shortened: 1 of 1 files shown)


seanmadsen:
@totten In our docs style guide we [say](https://docs.civicrm.org/dev/en/latest/documentation/style-guide/#parts-chapters-sections)

> Each chapter name should be short enough to fit nicely in the menu, but also long enough to stand on its own to a reasonable extent. The titles set here are used in the navigation menu and the page title that displays in the browser tab. The guide will be more usable if the reader sees two tabs titled "Using Hooks" and "API Usage" instead of "Usage" and "Usage".

So...

I like these changes because they still stand on their own:

* `s/Framework Reference/Framework/`
* `s/Development Standards/Standards/`

But with the other 5 changes, I worry that the page titles don't stand on their own enough. 

Aside: I wish that we could set the menu title separately from the page `<title>`. I've tried to find a way to do this in MkDocs but come up short. 

What do you think? Are you okay modifying this PR to only make the two changes above? I'd be okay merging it in that case. Or if you can make more of an argument for the other changes, I'm open to those as well but wary that we'd be sacrificing the `<title>` element (which helps with multi-tab usability and SEO).

totten:
Well, I'll take incremental progress on `Framework` and `Standards`. ;)

It's a good point about SEO and `<title>`. My gut still leans toward simpler titles - but it's not a really strong opinion. Either way is rational.

FWIW, with the longer titles, the incongruity between "(Coding,PHP,Database,Review) Standards" and "Javascript Reference" is a little grating, so I may toss in a little normalization there.

seanmadsen:
Looks good. I like `s/Javascript Reference/Javascript Standards/`.

PR #439, opened 2017-11-03
Restructure assetBuilder  example

aydun:
Restructure slightly based on Tim's comment: https://github.com/civicrm/civicrm-dev-docs/pull/425#issuecomment-338253390


totten:
I like it!

Made a few more tweaks (eg `asset-builder.md` went back/forth between `getUrl()` and `getURL()` -- this works because PHP methods are case-insensitive, but I put in `getUrl()` to match the capitalization in `AssetBuilder.php`.)

aydun:
Thanks - good updates.

PR #441, opened 2017-11-08
standards/review.md - Add templates. Tweak codes.

totten:
This adds the review templates. In fact, it adds more than one.

Why?  Based on survey results, the verbose(ish) multiple choice format
seemed to resonate most broadly.  However, based on some initial practical
usage, the more minimalist format felt easier to work with on an on-going
basis.

I tried to bridge this gap by:

 * Recommending the verbose(ish) format for new reviewers.
 * Recommending the abridged format for more experienced reviewers.


seanmadsen:
I like it. I added some small adjustments.

PR #442, opened 2017-11-09
standard/review - Consistent order. Use comments for instructions. Emphasize r-run prose.

totten:
Changes:
* Sort criteria in the same order across all three templates.
* For meta-instructions, use the Markdown/HTML comment notation.
* For `r-run`, we really don't want boilerplate. Remove the boilerplate.
* Prefer the DEL template over MC. DEL produces more readable output.


seanmadsen:
Cool. Yep. This all makes sense :white_check_mark:

PR #447, opened 2017-11-13
best practices for seting PATH

ginkgomzd:


seanmadsen:
Thanks Michael! I made some small adjustments to the markdown, most importantly to add the [necessary](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#admonition-syntax) double quotes around the custom admonition title.

totten:
@ginkgomzd Maybe you should publish a single-purpose web site like `bashpath.org` or `shellpath.net` or `path.readthedocs.org` dedicated solely to the instruction of "How do I add commands on the PATH?" ;)

I'm just kidding.

Well, half-kidding. There's a lot of CLI tools. But all of them have a sort of obligation to provide installation docs (because so many consumers skipped day 2 of the "Intro to Shell" class where this would be explained), but really it's redundant for all of them to do so, and all of the environments (Ubuntu, OSX, Windows, etc) do it a little different.

ginkgomzd:
@totten
It really is a testament to StackOverflow for making this kind of information readily available
I can't believe how many years I went without being clear on the concept of a login shell and the many different bash config files. Maybe if you go to unix sysadmin class, they really do cover this on day 2, but it does seem like this has long been assumed to be only relevant to sysadmins and not made available in luzer docs.

PR #454, opened 2017-12-05
Expand definition of "Non-Functional Change". Add examples and counter-examples.

totten:
If I disagree with using the "NFC" label, I'd like to be able to quickly
reference a link that explains the rationale and provides examples.


seanmadsen:
Cool! Great examples. Very helpful to have this level of clarity.

PR #466, opened 2018-01-03
CRM-21020 add additional params and example

lcdservices:


seamuslee001:
Thanks Brian, merging

PR #474, opened 2018-01-31
Review Standards - Change `r-jira` to `r-explain`

totten:
In keeping with the discussions about JIRA, Gitlab, and issue management
(e.g.  https://lab.civicrm.org/infrastructure/ops/issues/817), this change
would revise the "Review Standards".  Instead of specifically requiring
*a JIRA issue* for every PR, it requires *a public explanation* (which can be
provided by JIRA, Gitlab, etc).

Note: This change tries to be the smallest reasonable alteration to the
review criteria that accommodates Gitlab.  However, perhaps we could make it
better -- e.g.  by providing examples of good and bad explanations?

Related: https://github.com/civicrm/civicrm-core/blob/master/.github/PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/docs/standards/review.md
+++ b/docs/standards/review.md
@@ -28,7 +28,12 @@ If you were a site-builder reading the PR-log/release-notes and drilled into thi
 
 It is strongly encouraged that PR's include URLs/hyperlinks for any explanatory material (when available) -- such as a [JIRA issue](/tools/issue-tracking.md#jira), [Gitlab issue](http://lab.civicrm.org/), [StackExchange question](https://civicrm.stackexchange.com/), related PR, or [Mattermost chat](https://chat.civicrm.org). However, hyperlinks are not a substitute for a description. The PR should still have a description.
 
-Exception: [NFC](/tools/git.md#nfc) and [WIP](/tools/git.md#wip) PRs may not need a detailed explanation.
+PR descriptions should generally follow the [pull-request template](https://github.com/civicrm/civicrm-core/blob/master/.github/PULL_REQUEST_TEMPLATE.md), although this could be waived if another structure is more expressive.
+
+__Exception__: 
+
+* [WIP](/tools/git.md#wip) PRs do not need a detailed explanation until they're ready for earnest review.
+* Genuine [NFC](/tools/git.md#nfc) PRs do not need a detailed explanation. 
 
 ### Test results {:#r-test}
 
--- a/docs/standards/review.md
+++ b/docs/standards/review.md
@@ -26,7 +26,7 @@ Ensure the PR has an adequate explanation.
 
 If you were a site-builder reading the PR-log/release-notes and drilled into this PR, would you understand the description? If you were debugging a problem and traced the change back to this PR, would you understand why the change was made?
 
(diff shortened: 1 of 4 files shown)


agh1:
Continuing the conversation a bit from https://lab.civicrm.org/infrastructure/ops/issues/817, I think that you touch on a couple of key things that will be important to know about a PR: what issue it's trying to resolve, and what solution the change is making.

The additional thing in my mind that JIRA has provided is linkages:

- concurrent similar problems
- past attempts at solving the issue
- past changes that may have introduced unwanted behavior
- a larger epic that this may be a part of
- oustanding issues that follow the change
- categories the issue may belong to
- other related issues

totten:
I agree it's quite useful link to past attempts, related changes, etc.

Personally (and this reflects my own workflows more than an objective truth), linkages in JIRA usually show up as a 2nd or 3rd level thing. I *start out* reading a PR, chat, email, or StackExchange question -- and then eventually navigate through a JIRA issue mid-way to one of the others. So it leaves me *feeling* like JIRA is just indirection. For example:

* If the problem-identification discussion *really* happened in StackExchange, then I'd rather see the StackExchange link in the PR. (It feels silly to link from the PR to the JIRA issue, and then from JIRA issue to the StackExchange question.)
* If a PR implements one step from a bigger project-plan described by a wiki page, then I'd prefer the PR link directly to the relevant heading in the wiki page rather than some intermediate JIRA issue. 

However, this is entirely subjective -- your perception is going to depend on perspective (which skills/tasks/deliverables/workflows you typically engage) and happenstance (i.e. the particular issues you've been involved with).

--

I don't want to dis on JIRA -- I think its quality/experience for managing links is par for the course. In practice, I tend to see linkages in JIRA as follow:

* In issue-metadata, there are formal links to Github PRs.
* In issue-metadata *for waterfall projects* (a batch of preplanned issues), there are formal links.
* In comments/descriptions, there are issue-codes (like `CRM-11111`).
* In comments/descriptions, there are pasted URLs.

The only JIRA-ism that I might miss are issue-codes. But those are bitter-sweet... because they only work well if you're doing *everything* in JIRA. If you use other media (Mattermost chat,  git-commit messages, source-code comments, Github comments, email, StackExchange, etc), then `CRM-11111` is a little clumsy/opaque. And we do (in fact) use all of those things. I'd often rather see URLs -- they work in all those media.

agh1:
I wasn't suggesting that JIRA should be retained for linkages; I just want to make sure that linkages get made.  Linking to the "real" conversation in StackExchange or Mattermost or whatever would be ideal, but there's nothing stopping people from doing that already.

In contrast, we will be losing the automated linkages that are currently in place providing at least a connection among PRs referring to the same issue.  Somehow we'll need to insist that people paste those in.  (Without a shared issue ID, I don't think it'll be feasible to automate something like that.)

totten:
@agh1 I imagine a lot of this comes down to the judgment of the author/reviewer, but... I've updated the language to put more emphasis on the value of links.

totten:
Thank you, @seanmadsen.  

I'm inclined to merge for it now -- e.g. we had feedback from you and @agh1; we flagged up for a bit more visibility on the dev list; and it's getting a bit confusing to have `r-jira` around when we're seeing more things in Gitlab.  And we can certainly revise if there's more feedback.

Let's do it!

PR #477, opened 2018-02-08
dev/translation#2: update ts() example to E::ts().

mlutfy:
Since civix 17.08, we can use `E::ts("Hello")` instead of `ts("Hello", ['domain' => 'org.example.myextension'])`.


seamuslee001:
Changes look good to me make sense, There still seems to be enough about the legacy behavior in case people need it

Merging

PR #478, opened 2018-02-10
Update hook_civicrm_container.md

jaapjansma:
The civicrm core container does not extend the symfony ContainerBuilder class. There for remove it here otherwise the hook is not going to work.


seamuslee001:
Does this look right to you @totten ?

jaapjansma:
I am not completly sure either @seamuslee001

jaapjansma:
I have checked it myself and the docs are right. I was not using the hook_civicrm_container rightly.

totten:
Glad that got sorted. :)

PR #483, opened 2018-02-16
Describe `is_public` and `is_ssl` on the Routing page

agh1:
This adds descriptions for a couple of the XML elements for defining paths.  It also fixes alphabetization errors on the page.


totten:
Looks good to me.

PR #484, opened 2018-02-22
hook_civicrm_summaryActions.md - Fix formatting

JO0st:
added spaces so the last code block gets recognised as code


totten:
Thanks, @JO0st !

PR #487, opened 2018-02-28
Fix $this-> to self:: for an static property

yaazkal:
This will prevent an error like the described if a person is following this tutorial.

```
Notice: Accessing static property CRM_Core_Payment_CUSTOM::$_mode as non static in CRM_Core_Payment_CUSTOM->__construct() 
```

Where `CUSTOM` is a placeholder for the procesor the dev is making.

### DIFF
--- a/docs/extensions/payment-processors/create.md
+++ b/docs/extensions/payment-processors/create.md
@@ -113,7 +113,7 @@ class edu_ucmerced_payment_ucmpaymentcollection extends CRM_Core_Payment {
    * @return void
    */
   function __construct( $mode, &$paymentProcessor ) {
-    $this->_mode             = $mode;
+    self::$_mode             = $mode;
     $this->_paymentProcessor = $paymentProcessor;
     $this->_processorName    = ts('UC Merced Payment Collection');
   }


totten:
This property appears to be declared as `static` within the example, so the change seems more internally consistent. (Note: There are two more spots in this file which mention `$this->_mode`... but I haven't read closely enough to see if they're the same thing.)

@eileenmcnaughton , just out of curiosity -- does it seem like better policy for a payment processor class to have `$_mode` as `static` or non-`static` property?

yaazkal:
@totten @eileenmcnaughton please let me know what is the policy, so I can even commit more to this PR "fixing" the other variable calls or just removing the `static`.

I'm reveiwing it since at the time I'm also building a module for a local payment processor in my country (I'll also send it to be published on civicrm.org)

Regards !

eileenmcnaughton:
I have started to work on a PR to properly update that  - the long & short of it is that var should not be in the example at all

https://github.com/civicrm/civicrm-dev-docs/pull/488

yaazkal:
@eileenmcnaughton ok, thanks a lot!

I'm closing this PR then.

PR #496, opened 2018-03-21
tools/jenkins.md - Add tips for reviewing console output

totten:
@seamuslee001 @seanmadsen - I was looking for somewhere to put tips on how to read/search the Jenkins console output when reviewing `r-test`. In particular, searching for `EXITCODE` can be helpful.

After searching the docs for `console out`, they seemed to fit in the `tools/jenkins.md` discussion. 

Thematically, there's some overlap between `tools/jenkins.md` and `core/pr-review.md`. It seems like a pre-existing issue between these pages, but a separate improvement might be to link these bits of content better?


seamuslee001:
@totten i think this is good, the only other things is it might be good to write about searching for a PHP Fatal error or similar in the console log as well

totten:
@seamuslee001 I used to search on PHP Fatal, but then I realized that EXITCODE is more general -- basically, anything that produces a PHP fatal should also non-zero EXITCODE. But EXITCODE will match additional things that don't show up as a PHP fatal.

seanmadsen:
Cool. Thanks @totten. This seems like useful stuff! 

I'm gonna go ahead and merge now but if we want to add more info about "PHP Fatal" we can do that in a separate PR.

PR #497, opened 2018-03-22
Tokens/Mailings - Split in two files. Multiple cleanups.

totten:
Including the token documentation under CiviMail is misleading -- the docs talk about CiviMail because, *historically*, much of this grew out of CiviMail. However, *today*, tokens are used in several more contexts. Also, in `civicrm-core`, `TokenProcessor` is *not* used by CiviMail. (You have to install FlexMailer to use `TokenProcessor` on mailings... and you may also have to enable an experimental setting...)

### DIFF
--- a/docs/framework/civimail.md
+++ b/docs/framework/civimail.md
@@ -18,7 +18,7 @@ Note that a contact may only have one subscription record for the group, so the 
 
 ## Tokens {:#tokens}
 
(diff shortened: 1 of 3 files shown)


seamuslee001:
I looked over this and i am happy with the changes.  @michaelmcandrew any thoughts?

totten:
Thanks, @seamuslee001. Will go ahead and merge to prevent staleness - but further feedback from Michael is of course appreciated.

PR #503, opened 2018-03-23
Docs style guide changes and improvements

seanmadsen:
Three distinct changes to the style guide, as described in each commit. 

I'd like some feedback from these folks before continuing: 

* @michaelmcandrew 
* @joannechester

### DIFF
--- a/docs/documentation/style-guide.md
+++ b/docs/documentation/style-guide.md
@@ -11,15 +11,16 @@ uphold to ensure all guides maintain this high level of quality.
 
 To streamline communication about documentation issues, we define the following terms:
 
--   "part" - folder
--   "chapter" - file (in markdown), also one web page with a given URL
--   "section" - heading within the page
+-   "Chapter" - folder
+-   "Sub-chapter" - a folder within a folder
+-   "Page" - file (in markdown)
+-   "Section" - heading within the page
 
-### Chapter names
+### Page names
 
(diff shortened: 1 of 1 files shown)


joannechester:
I agree that 4 levels of menu can be needed for some topics.  I was thinking I needed that extra level when planning the reorganisation of the email chapter. It makes no sense to have A/B testing at the second menu level as it is only relevant to CiviMail. so we need Email - Mass mailings using CiviMail - A/B testing 

I am not so keen on the nomenclature for the 4 levels.  To me "page" doesn't fit.  I think of a page as the place the information is displayed, not as a part of the content that is being displayed.  This is obviously true for a physical book, but I think it also applies online where a page can have a heading, side bars etc.  The content is just one part of it. 

However, having spent a couple of hours trawling on this topic, I can't find a definitive alternative. As it fits will how things are already described in the [documentation guide](https://docs.civicrm.org/dev/en/latest/documentation/markdown/#internal-url-standards) so I am OK with it.

dvhirst:
Per discussion, this seems very sensible to me.  Glad to have a definitive 1 - 1 relationship specified.  Don

seanmadsen:
Merging now since we've had a few opinions and discussed it at the meeting.

https://lab.civicrm.org/documentation/docs/blob/master/meetings/notes/2018-04-03.md#style-guide

PR #505, opened 2018-03-23
Update the list of Pre and Post EntityTypes after doing…

seamuslee001:
Used the following to find the entites
`grep -ir 'CRM_Utils_HooK::pre' CRM/*` and `grep -ir 'CRM_Utils_HooK::post' CRM/*`

@seanmadsen can you review this if your happy we can merge


seanmadsen:
Nice!

PR #507, opened 2018-03-24
toosl/git.md - Describe base branches in more detail

totten:
There are usually a few branches to choose from when picking a base-branch.
This breaks out a table to differentiate them.


seamuslee001:
This is good Tim, I converted the table into a markdown format table. Merging.

PR #509, opened 2018-03-24
Resolve #491 Reference hook_civicrm_coreResourceList in the advanced …

### LINKED ISSUE #491: Reference hook_civicrm_coreResourceList in https://docs.civicrm.org/dev/en/latest/framework/resources/
see https://chat.civicrm.org/civicrm/pl/tgn1siojhbfgbkwrsz3tgduxjr and the following messages.

It would be good to highlight hook_civicrm_coreResourceList as a global way to add resources.

seamuslee001:
…options in Resource framework reference

ping @seanmadsen @michaelmcandrew. Does this cover what we need?

### DIFF
--- a/docs/framework/resources.md
+++ b/docs/framework/resources.md
@@ -91,3 +91,5 @@ For Smarty APIs (`{crmScript}` and `{crmStyle}`), the optional weight and region
 {crmScript ext=com.example.foo file=bar.js weight=10 region=page-footer}
 {crmStyle url="http://example.com/bar.css" weight=10 region=page-footer}
 ```
+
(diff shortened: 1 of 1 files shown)


seanmadsen:
I'm gonna wait to see if @michaelmcandrew can review this one.

seanmadsen:
Well it's been a while with this sitting here so I'm inclined to get this in even though I can't vouch for the accuracy of the content without putting more time into it. @michaelmcandrew if you can give this a second look that'd be nice, but I'm going to go ahead with a merge here since this seems pretty low-risk.

PR #511, opened 2018-03-26
Corrected civix generate:custom-xml

andrewpthompson:
Changed `civix generate:custom-data` to `civix generate:custom-xml` as the former is incorrect. Also corrected grammar in subheading "Extending a subtype".


seanmadsen:
Thanks!

Issue #389, opened 2017-10-07
Migrate testing content from Testapalooza


seanmadsen:
Good content here from @totten that should probably be migrated into the Dev Guide: 

https://github.com/civicrm/org.civicrm.testapalooza/blob/master/civi-test.md

seanmadsen:
Hey @totten I just noticed there is still a bunch of content here: https://github.com/civicrm/org.civicrm.testapalooza/blob/master/civi-test.md

Want to delete that and replace it with text that directs readers to https://docs.civicrm.org/dev/en/latest/testing/phpunit/ ?

I noticed this because I happened to run `civix generate:test` and read the comments in the file which directed me to testapalooza.

seanmadsen:
Actually, what is the function of the Testapalooza project anyway? Is it still relevant? Can we just move *all* of that content into the Dev Guide?

totten:
> Actually, what is the function of the Testapalooza project anyway? 

Testapalooza was an initiative to re-design the test system -- to make it less of a framework/monolith (hard-wired into civicrm-core) and more of a utility/library (that can be incorporated into other projects). It produced a git repo (with docs/examples) and several utilities (`cv`, `CiviTestListener`, `Civi\Test`).

> Can we just move all of that content into the Dev Guide?

I think most of it has been -- either in `testing/index.md` or `testing/phpunit.md`.

Maybe what you're really saying is, "Let's delete the old version"? I'm not so eager on that. 

*Analytically*, most of the substance has been reproduced. But there's still *something* that doesn't come through. As if... the Dev Guide were written at a 9th-grade reading-level, and the Testapalooza docs were written at a 12th-grade reading-level.? Or as if... the audience or perspective is different? Or maybe it's my own sentamentalilty? Or maybe it's a fear that bits of information will be removed from the Dev Guide -- and I like having Testapalooza docs as a fallback (b/c they're less filtered -- they map closer to how I really think about the problem-space). Whatever it is, I can't quite put my finger on it.

The best analogy I can think of... we often publish blogs when some milestone is reached. We don't delete the blogs when technical details change. However, this doesn't preclude us from keeping "living documentation" about the most common use-cases/perspectives.

PR #512, opened 2018-03-26
tests/phpunit.md - Document multiple helper classes. Provide specific examples of writing tests.

totten:
This is a significant revision of the PHPUnit documentation. This specifically:

* Adds concrete examples of writing tests in various styles (`CiviUnitTestCase`, `CiviEndToEndTestCase`, `civix`)
* Adds reference documentation for classes/traits/interfaces provided by Civi's test library
* Imports key information from Testapalooza docs (#389)

### DIFF
--- a/docs/testing/phpunit.md
+++ b/docs/testing/phpunit.md
@@ -11,7 +11,7 @@ ensuring that the `Contact.create` API actually creates a contact.
 PHPUnit is a command-line tool, but the command name varies depending on how it was installed. For example:
 
 * In [buildkit](/tools/buildkit.md), this command is named `phpunit4`.
-* In other environments, it might be `phpunit` or `phpunit.phar`.
+* In other environments, it might be `phpunit` or `phpunit.phar` or `phpunit.bat`.
 
 For the following examples, we'll use `phpunit4`.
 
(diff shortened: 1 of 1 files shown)


docs/testing/phpunit.md
```diff
@@ -123,52 +127,297 @@ $ env CIVICRM_UF=UnitTests phpunit4 ./tests/phpunit/CRM/Core/RegionTest.php
...
+    </whitelist>
+  </filter>
+</phpunit>
+```
+
+At a minimum, the `bootstrap.php` script should register CiviCRM's class-loader, e.g.
+
+```php
+eval(`cv php:boot --level=classloader`);
+```
```
seamuslee001: @totten just wondering here if its better to in someways not use `cv` as an example reason being is that not everyone may have cv installed on the server and i would have thought we would want the booting to be able to be done without the need for additional tools like cv

seanmadsen:
Wow! This is some substantial work. Thanks @totten! I just skimmed the diff and everything looked okay to me, formatting-wise. I'd feel comfortable merging this as-is once Seamus' issue is resolved.

docs/testing/phpunit.md
totten: (1) Tests in `civicrm-core` and `civix` already depend on `cv` in exactly this way. The aim of a "`$foo`-from-scratch" is to build up toward real/typical code.

(2) This line is the original raison d'etre of `cv` -- there isn't a simpler alternative (at least, none near as good). The basic problem is that the cumulative file-structure of CMS+Civi is f'ing wild. `cv php:boot` hides this wildness by (a) taking advantage of automatic `PATH` searching and (b) having a default that works in most environments.

(3) For rhetorical purposes, I guess we could present "`$foo`-from-scratch" a bit differently -- e.g. start out with a simpler `bootstrap.php`, then talk about all the ways in which that's broken/incomplete, and then present `cv php:boot`.  I've pushed up some changes like this.

totten:
Thanks, @seamuslee001 @seanmadsen. I've updated the "from scratch" example to build up a bit more gradually (easing into the `cv` bit) -- and also made a bunch of copy-edits.

seamuslee001:
Thanks Tim looks good to me :-)

PR #513, opened 2018-03-27
Say "internal link" instead of "internal URL"

seanmadsen:
@joannechester pointed out that I've been using "URL" to refer to things that are not URLs! Wow! I never knew! 

Per https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#URI_references it seems that something like `/foo/bar.md` isn't technically a "URL". So to be more accurate with our language, this PR changes the meta docs to refer to internal "links" instead.

Note that I intentionally preserved the anchor `#internal-url-standards` to avoid broken references.

What do you think @joannechester? Is this the sort of change you had in mind?

### DIFF
--- a/docs/documentation/markdown.md
+++ b/docs/documentation/markdown.md
@@ -33,10 +33,10 @@ To maintain some consistency and peace of mind for documentation content editors
 * **Unordered lists:** use `*` to delimiters.
 * **Headings:** use hashes like `## Heading 2`.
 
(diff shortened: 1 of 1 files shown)


joannechester:
Yes, this looks good to me.  OK to merge as far as I am concerned.

seanmadsen:
Thanks @joannechester :slightly_smiling_face:

PR #515, opened 2018-04-02
Document `civix generate:entity-boilerplate`

seanmadsen:
Awesome work by @michaelmcandrew to [add the `civix generate:entity-boilerplate` command](https://github.com/totten/civix/pull/124)!!

I think we should direct people to this new command now. I think even if it's still "experimental" it's still the best thing we have, and it'd be great to get more people using it in order to improve it!

@michaelmcandrew would you care to review this PR?


michaelmcandrew:
hey @seanmadsen - thanks for updating. it looks about right to me :)

PR #517, opened 2018-04-03
trivial: remove layer of quotes

michaelmcandrew:
Two arrows were being interpreted as two layers of quotes. I removed one of them. Then I thought, actually, these are not quotes. Maybe we should remove move. And maybe just having them on their on paragraph without any quotes is good enough. I am shying away from using boxes for this since to avoid box fatigue but would be interested in some opinions on the house style.

### DIFF
--- a/docs/extensions/index.md
+++ b/docs/extensions/index.md
@@ -33,9 +33,9 @@ to help decide.
 ### 1. Install civix
 Some tasks in the process of writing an extension require boilerplate code. To reduce the amount of work required to find, understand, and adapt the boilerplate code, one should install the CiviCRM extension builder, civix. Civix is a command-line tool which generates code for some common development tasks.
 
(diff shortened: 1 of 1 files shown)


seanmadsen:
I think that content needs a bit of cleaning and re-working.

* I agree that one level of block quote is better than two. But I'd suggest removing those block quotes altogether and working the links into the sentences. There's nothing *so* special about those links to make them stand out I don't think.
* Also, the "civix" link should be an internal link to the civix page (instead of an external link to the civix project).
* Also links with text like "read on" are a huge pet peeve of mine because they make the page harder to skim. I shouldn't have to read a sentence in order to understand where a link will take me. So that link should just be on the text "info.xml".

seanmadsen:
I'm going ahead with a merge here. As I said above, that content really needs some more substantial cleaning, but this PR certainly doesn't hurt!

PR #521, opened 2018-04-19
Add docs for hook_civicrm_alterLocationMergeData

JKingsnorth:
There wasn't any documentation for hook_civicrm_alterLocationMergeData, so I thought we should create something! It was introduced in 4.7.10.

@eileenmcnaughton is there anything you think we should add to this?

It would be good to get something basic up, which we can embellish later if needed.

I'm not sure if I've edited this in all the right places, let me know if I need to make any changes.

### DIFF
--- a/docs/hooks/hook_civicrm_alterLocationMergeData.md
+++ b/docs/hooks/hook_civicrm_alterLocationMergeData.md
@@ -1,0 +1,44 @@
+# hook_civicrm_alterLocationMergeData
+
+## Summary
+
+This hook allows you to alter the location information that will be moved from
+the duplicate contact to the master contact.
+
+## Availability
+
+This hook was first available in CiviCRM 4.7.10.
+
+## Definition
+
+```php
+hook_civicrm_alterLocationMergeData(&$blocksDAO, $mainId, $otherId, $migrationInfo)
+```
+
+## Parameters
+
+   * array $blocksDAO: Array of location DAO to be saved. These are arrays in 2 keys 'update' & 'delete'.
+   * int $mainId: Contact ID of the contact that survives the merge.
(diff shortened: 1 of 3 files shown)


eileenmcnaughton:
Looks good to me @seanmadsen can you merge?

seamuslee001:
done @eileenmcnaughton

eileenmcnaughton:
thanks @seamuslee001

JKingsnorth:
Thanks all!

seanmadsen:
Thanks @JKingsnorth 

I made a couple follow-up changes:

* [Improve hook formatting ](https://github.com/civicrm/civicrm-dev-docs/commit/abab6fca7948ebbb18746400946e7bb2e7561ec4)

* [Regenerate hooks summary page ](https://github.com/civicrm/civicrm-dev-docs/commit/e063f0fd5c6d8677e4622ff2eb69a03c862b0fce)
    * Note to commiters like @eileenmcnaughton and @seamuslee001 that we need to perform this step manually whenever we add a new hook or modify a hook description. Perhaps someday we'll evolve to a more sophisticated build process that will do this automatically (which is why Michael Devery chose to write the script in Python, so it can be incorporated into an MkDocs extension). But for now, this is the only auto-generated content in any of our guides, so its changes just get committed into git. I'm open to suggestions about how to improve this (as long as they don't require lots of work :grin: ).

eileenmcnaughton:
Thanks all - I had been delaying reading all these notifications because I was looking for time to respond to the request to document this. Just now I finally wrote it on my notebook to-do list & opened it in order to be able to clear out other to-do lists - imagine my happy surprise to discover you had all already done it!

PR #524, opened 2018-04-20
Create hook_civicrm_alterAngular.md

colemanw:
See https://github.com/civicrm/civicrm-core/pull/10085
@totten I've added docs for this hook.

### DIFF
--- a/docs/framework/angular/changeset.md
+++ b/docs/framework/angular/changeset.md
@@ -19,7 +19,11 @@ independent deployments with independent configurations).
 
 ...
 
-## tldr
+## Hooks
+
+* **[hook_civicrm_alterAngular](/hooks/hook_civicrm_alterAngular.md)** - Alter the definition of some Angular HTML partials.
+
+## Example
 
 ```php
 function mailwords_civicrm_alterAngular(\Civi\Angular\Manager $angular) {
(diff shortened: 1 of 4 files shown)


totten:
Good idea adding a page under "Hooks".

FWIW, there's also this page: https://docs.civicrm.org/dev/en/latest/framework/angular/changeset/ 

At the moment, the pages are pretty similar, but the ambition in `framework/angular/changeset` was to also talk about the workflow with the helper commands

```
cv ang:html:list
cv ang:html:show <file>
cv ang:html:show <file> --diff
```

totten:
@colemanw I suggest hyperlinking between the pages and then merging.

seanmadsen:
@colemanw also run `./bin/hook-summary.py` to re-generate the hooks summary page based on the content of your hook file.

colemanw:
@seanmadsen I've updated the pages per @totten's suggestion. Could you please check to see if I did the hyperlinks correctly?
I tried running the python script but it didn't do anything. It ran but gave no output and did not update the summary.md file :/ Might be a problem with python on my system?

seanmadsen:
@colemanw 

> the python script but it didn't do anything

That's because the hook needed to be added to the menu (`mkdocs.yml`) first. (The python script reads from the menu file to generate the page.)

I modified the menu to add the hook. I also tweaked the hook description and regenerated the summary page. I think we're good to go here, so I'm merging.

PR #525, opened 2018-04-23
Added and improved civilint examples

michaelmcandrew:
I think these two are more useful examples. Could be just me :)


seanmadsen:
Great!

PR #532, opened 2018-05-17
Update docs to reflect semver changes.

MikeyMJCO:
Updating docs to reflect changes made here: https://lab.civicrm.org/dev/release/issues/1#note_2843


totten:
Thanks, @MikeyMJCO. This is good to document. 

I'd forgotten there was a revision table at the top of the document -- glad you found both spots.

I pushed up a change - mainly replacing the phrase "to bridge this change" (07240a6) because we're in a new context.

PR #535, opened 2018-06-10
Update debugging.md

MikeyMJCO:
Clarified when the backtrace is displayed, i.e. calling &backtrace=1 does not trigger an immediate backtrace unless an error occurs.


totten:
Thanks, @MikeyMJCO  - seems like a good idea to add that info. 👍

PR #536, opened 2018-06-10
Update upgrade documentation to mention how to upgrade message templates

eileenmcnaughton:


docs/framework/upgrade.md
```diff
@@ -38,6 +38,9 @@ Each class in the `CRM_Upgrade_Incremental_php` corresponds to a major release o
 | `php public function upgrade_X_Y_Z() ` | Perform any changes in a specific queue order for version X.Y.Z. During this function you *MUST* call `php $upgrade->processSQL($rev)`. Without calling this function the X.Y.Z.sql.tpl file will not be executed |
 | `php public function setPostUpgradeMessage() ` | Called after each incremental upgrade. Any messages generated by this function will be displayed to site-administrators *after* the upgrade has completed. (This is called several times – once for each incremental upgrade.) |
 
+## Upgrading message templates
+Message templates that need upgrading should be declared - currently under CRM_Upgrade_Incremental_MessageTemplates::getTemplateUpdates although this may be defined. The upgrade script will pick them up from there (5.4+)
+
```
seamuslee001: @eileenmcnaughton should we give a definition of the format the array needs to be in?

eileenmcnaughton: @seamuslee001 possibly - except I feel it might mature a bit

aydun:
It would be useful to document how to determine the right version to use in upgrade steps.

totten:
Thanks @eileenmcnaughton. I've simplified the language a little bit and will merge.

@aydun, that's a good point about determining the version. To my eye, it merits a different subsection/PR for this doc -- because the choice of version is important for (a) message-template upgrade steps as well as (b) incremental SQL-based upgrade steps and (c) incremental PHP-based upgrade steps.

PR #537, opened 2018-06-12
Update hook_civicrm_post.md

kainuk:
See the https://lab.civicrm.org/dev/core/raw/master/CRM/Contact/BAO/Contact.php line 1188

### DIFF
--- a/docs/hooks/hook_civicrm_post.md
+++ b/docs/hooks/hook_civicrm_post.md
@@ -31,8 +31,7 @@ hook_civicrm_post($op, $objectName, $objectId, &$objectRef)
     -   'create' : The CiviCRM object is created (or contacts are being added to a group)
     -   'edit' : The CiviCRM object is edited
     -   'delete' : The CiviCRM object is being deleted (or contacts are being removed from a group)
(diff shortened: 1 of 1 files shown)


totten:
@kainuk Nice catch.

Just to verbalize something... it wasn't clear to me if this patch (a) corrects a mistake in docs that were never accurate or (b) updates docs to match a change in the `hook_civicrm_post` contract. If there was actually a change in the contract, then the docs should indicate when that change happened. 

Trying to answer that... I grepped the [release-notes](https://github.com/civicrm/civicrm-core/tree/master/release-notes) and [changes.md](https://docs.civicrm.org/dev/en/latest/api/changes/) for `civicrm_post` and couldn't find any mention of this. I also spot-checked the code in a few releases (`5.3`, `5.0`, `4.7.20`, `4.7.10`, `4.7.0`, and the current `4.6`) to see if the symbol `'trash'` appeared to be emitted. Couldn't find anything that looked like hook-invocation.

So, based, on that, I'm guessing that this was really a correction for mistaken docs, and I'm going to merge. However, grep isn't perfect... and @kainuk, you may know more about this than me... so if this was actually a change in contract, please follow-up.

PR #540, opened 2018-06-27
Update hook_civicrm_selectWhereClause.md

michaelmcandrew:
Add a note to say that this hook will only be invoked for API calls if check_permissions is set to 1


seamuslee001:
Looks fine to me merging

PR #545, opened 2018-08-23
Fix incorrect sample filename

selwyntcy:


totten:
Thanks, @selwyntcy . I haven't read this page in a long time and don't immediately recognize the problem being fixed. Trying to read+summarize, it seems like:

* __Background__: There are supposed to be two files -- for a *class* and an *entry-point*.
* __Before the patch__: The short names of the two files were the same (i.e. `UCMPaymentCollectionIPN.php`, although maybe the folders could be different) -- and this was confusing.
* __After the patch__: With the patch, the names are clearly different (`UCMPaymentCollectionNotify.php` vs `UCMPaymentCollectionIPN.php`).

Have I understood correctly?

selwyntcy:
Hm. Both [https://github.com/civicrm/civicrm-dev-docs/blame/master/docs/extensions/payment-processors/create.md#L302](url) and [https://github.com/civicrm/civicrm-dev-docs/blame/master/docs/extensions/payment-processors/create.md#L308](url) mention that there should be a file named `UCMPaymentCollectionIPN.php`. However, the filename above such sample code is incorrectly written as `UCMPaymentCollectionNotify.php`, the same as that of the previous sample code. I guess it was a copy & paste error.

mattwire:
@totten the whole page could do with an update.  But this PR is an improvement and should be merged.

totten:
Thanks, @mattwire

PR #548, opened 2018-10-07
review.md - Split review standards into two buckets, "General" vs "Developer"

totten:
This was raised during the Bamford discussion about PR review -- it should make it easier for less sophisticated reviewers to recognize the parts where their feedback can help.


mattwire:
@clairew @joannechester Is this ok to merge?

totten:
@mattwire, thanks for doing the follow-up pings.

I think we've given this a reasonable comment period and can go ahead with merging.

PR #555, opened 2018-11-05
Update setting metadata information

eileenmcnaughton:
Note working through this threw up some ambiguity about help_text vs help which I'll dig into as a follow up. I think there is further text on the page I need to review too the section about the settings form


eileenmcnaughton:
OK - the help text is not used anywhere - ie. the text defined this way does not show up anywhere. I think it's OK to leave it as is - I guess the question is whether the new 'help' => 'id' methodology (used on entity forms) is good or if we should try to do as assign to help text

MegaphoneJon:
This appears to be superseded by #557.

PR #561, opened 2018-11-22
Add Civix example

eileenmcnaughton:


totten:
Thanks, @eileenmcnaughton. 👍 Agree that's better example for typical usage of `navigationMenu`.

PR #464, opened 2017-12-18
CRM-21460 Add job execution hooks

rthouvenin:
This is the developer documentation for CRM-21460 ([PR#11377](https://github.com/civicrm/civicrm-core/pull/11337)). Adds 2 new hooks around scheduled job execution.

PR #562, opened 2018-11-28
Fix missing reference to pre/postJob hook files.

MikeyMJCO:


MikeyMJCO:
This PR fixes an issue with #464 which is currently preventing the Dev docs from publishing.

PR #563, opened 2018-11-29
Add "REF" (Refactor) as prefix for PR/git subjects

totten:
This documents "REF" (refactor) as an alternative to "NFC" (non-functional commit).

When coining "NFC", I personally expected that it would not include refactorings -- because refactorings change how the code executes. But other folks have found it appealing to mark refactorings as "NFC". I think there's a kernel of validity in both these expectations/behaviors. I'm hoping we can resolve that minor tension by adding another category specifically for refactoring.

As a someone reading PRs and commits, you would be able to use these mental heuristics:

* "NFC" means "I need to skim the PR and understand it on the author's original terms. If there's a small mistake -- such as a new typo -- it's not going to break anything."
* "REF" means "I should check if this change is a safe refactoring -- aka 'behavior-preserving transformation'."

See also: https://github.com/civicrm/civicrm-core/pull/13167#issuecomment-442651462


totten:
Merging based on linked discussion with Eileen.

PR #564, opened 2018-11-30
Under Hooks, switch the order of 'All Hooks' and 'Setup'

aydun:
Currently we have 'All Hooks' followed by each of the categories of hooks
so it looks like 'Setup' is a hook category.  Its content is closer to 'Using Hooks'

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -58,12 +58,12 @@ pages:
   - APIv3 Changes: api/changes.md
 - Hooks:
   - Using Hooks: hooks/index.md
-  - All Hooks: hooks/list.md
   - Setup:
     - Hooks with Symfony: hooks/setup/symfony.md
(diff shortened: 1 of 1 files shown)


totten:
Agree, I had the same confusion when I tried to navigate to the Symfony one. (And I wrote that page!) Wound up doing a Google search instead.

Merging because this feels better than the old menu and is a minimal change.

FWIW, I still think the word-choice here is confusing.

* "Using" is... OK for the content, although "Intro" or "Overview" feels more canonical to me.
* "Setup" feels misleading -- "Usage" or "Syntax" or "Bindings" make more sense to me.

@aydun, if you also think the word-choice is confusing, then I'd be happy to draft changes or review changes you draft. Or if you're OK with the word-choice, I can move on.

PR #565, opened 2018-11-30
Add in documentation of crmType and phpType setting now avaliable as …

seamuslee001:
…of CiviCRM 5.9 for Extensions

ping @totten 

Merged PR is here https://github.com/civicrm/civicrm-core/pull/13059


mattwire:
@seamuslee001 Is this still relevant? Is there an example extension that could be linked to?

seamuslee001:
@mattwire yes see here https://github.com/australiangreens/au.org.greens.civigeometry/blob/master/xml/schema/CRM/CiviGeometry/Geometry.xml#L87 we create a MYSQL column of type geometry but use the CRM type of CRM_Utils_Type::String you can see how that renders here https://github.com/australiangreens/au.org.greens.civigeometry/blob/master/CRM/CiviGeometry/DAO/Geometry.php#L202

PR #567, opened 2018-12-13
Restructure to better incorporate Symfony hooks

aydun:
Remove some old content
Generally refresh

@totten - this is a follow-up to your comment on #564  I've used "traditional method" - happy to change if you have better wording!


mkdocs.yml
```diff
@@ -57,12 +57,15 @@ pages:
...
-    - Hooks with WordPress: hooks/setup/wordpress.md
+  - Hooks Introduction: hooks/index.md
+  - Usage:
+    - Traditional Hooks: hooks/traditional.md
+    - Hooks with Symfony: hooks/symfony.md
+  - Legacy CMS Hooks:
+    - CMS Hooks: hooks/cms/cms.md
+    - Hooks with Joomla: hooks/cms/joomla.md
+    - Hooks with Drupal: hooks/cms/drupal.md
+    - Hooks with WordPress: hooks/cms/wordpress.md
```
totten: @aydun I like the symmetry that comes from moving some of those examples to a separate page. Other thoughts:

1. The words "legacy" and "traditional" are a bit loaded. Personally, I don't think of this as "legacy vs modern" or "old vs new"; I think of this as different packaging formats for different use-cases. (To wit: an add-on to "Sync CiviCRM Group<=>Drupal Role" is qualitatively different from an add-on to incorporate "Mosaico UI"; the differing requirements lead to different packaging choices.)

2. The phrase "Legacy CMS Hooks" sounds too similar to the subsequent phrases "Batch Hooks", "Case Hooks", "Database Hooks" -- i.e. this is about the syntax used with a CMS packaging format.

How about:

```
- Hooks Introduction: hooks/index.md
- Usage:
  - Hooks in Extensions: hooks/usage/extension.md
  - Hooks in Symfony: hooks/usage/symfony.md
  - Hooks in Drupal: hooks/usage/drupal.md
  - Hooks in Joomla: hooks/usage/joomla.md
  - Hooks in WordPress: hooks/usage/wordpress.md
- All Hooks
- Batch Hooks
- Case Hooks
...
```

Granted, there is still some subtlety that doesn't come through in the TOC (the "Symfony" notation works in any environment; the "Drupal" docs also apply to "Backdrop"), but IMHO it's better to convey that with a few words in the `index.md` (where you've a couple small examples/intro/hyperlink) rather than the TOC.

3. If we change names of the folders or files, then there also need to be updates in https://github.com/civicrm/civicrm-dev-docs/tree/master/redirects

aydun:
Hi @totten, I've restructured as you suggested.

However, I'm still not sure that "Hooks in Extensions" vs "Hooks in Symfony" quite works in that extensions can use either the Symfony or non-Symfony style.  I hear your point about 'traditional' but I think we're looking for a suitable name for the style I've labelled 'traditional' in the current iteration.   

Do you want to try rewording the 'Usage' section of `index.md`?  

Re 3) - Not sure quite how 'redirects' get used, but it's not obvious to me what if anything needs to change there.

mattwire:
@aydun @totten is this ok to merge now and improve later in subsequent PRs?

PR #570, opened 2018-12-21
Fleshed out validate_callback doc.

GinkgoFJG:


totten:
Yup, good additions. Merging.

FWIW, the callback can be expressed in a few more notations, although they may not all be useful in this context.

https://github.com/civicrm/civicrm-core/blob/5.9/Civi/Core/Resolver.php#L13-L24

PR #574, opened 2019-02-15
Correct function name

ray-wright:
I believe the getMenuKeyMax function is missing it's underscore inside the hook example.


mattwire:
I wonder if this "legacy" method should be removed altogether?

PR #578, opened 2019-02-15
don't include deprecated attribute in example

jmcclelland:
Many users will simply copy and paste the example settings file. However, this example contains an attribute that is deprecated.


mattwire:
Correct. ok to merge

Issue #579, opened 2019-02-15
more explanation if is_domain and is_contact for settings would be really helpful


jmcclelland:
The [settings reference](https://docs.civicrm.org/dev/en/latest/framework/setting/) is generally excellent. 

However, there are two attributes that I still don't quite understand: is_domain and is_contact.

It's not clear to me when creating settings for an extension when and how I should set those attributes.

totten:
TLDR: Set either `is_domain=1` XOR `is_contact=1`. 99% of usage would be `is_domain=1`.

Not sure how to update docs, but just to put out some more detailed interpretation/observation:

* Settings can be tracked on a "per-domain" xor "per-contact" basis.
* The choice of "per-domain" or "per-contact" is a static/up-front decision. Once you decide that a setting `foo` will be per-domain, then `foo` should always be per-domain. There is no hierarchy/precedence.
* You work with them a bit differently -- e.g.
    * Per-domain `Civi::settings()` vs per-contact `Civi::contactSettings()`.
    * Per-domain `$civicrm_setting['domain']` vs per-contact `$civicrm_setting['contact']`. (Note: Any other/unrecognized value is treated as `domain`.)
    * The form-generators are used with per-domain; I've never seen them used with per-contact.
* The model is somewhat redundant -- in `*.settings.php`, you set `is_domain` XOR `is_contact`. In table `civicrm_setting`, you  set `is_domain` XOR `contact_id`. (The system may not complain if you do other combinations... but it would be hard to understand/reason about.)
* In practice, 99% of the settings are per-domain. Grepping around, I could only find two per-contact settings -- `activity_tab_filter` in `civicrm-core`; `tutorials` in `org.civicrm.tutorial`. I don't think either of them are defined in `*.settings.php`.

seamuslee001:
The other thing is that you can in a multisite context set `is_domain = 'all'` and it will set the setting across every domain. This is useful if you were doing something like setting the prevNext Backend setting

jmcclelland:
Thank you! Immensely helpful and now that you have explained it, the answer is quite obvious. I've submitted a pull request that I hope clarifies it for others.

seamuslee001:
Closing this as PR #582 has been merged

PR #582, opened 2019-02-18
clarify meaning of is_contact and is_domain

jmcclelland:
See https://github.com/civicrm/civicrm-dev-docs/issues/579


mattwire:
Good to merge

PR #585, opened 2019-03-06
Add note about custom data on any creation

eileenmcnaughton:


eileenmcnaughton:
Hmm - here is an example of adding the option value https://github.com/eileenmcnaughton/org.wikimedia.relationshipblock/blob/master/CRM/Relationshipblock/Upgrader.php#L16

MegaphoneJon:
I'm going to merge this for now, but I still do intend to write something more definitive...eventually.

PR #589, opened 2019-03-20
Update hook_civicrm_aclWhereClause.md

artfulrobot:
Clarify that the ContactID refers to the user making the query - my reading had been that it was the contact ID of the contact that was being considered.


mattwire:
@colemanw Ok to merge

colemanw:
@mattwire does the hook documentor script overwrite manual changes to hook docs? And if so then should this actually be a PR agains core to update the docblock?

mattwire:
@colemanw No, unless there is another script.  This one https://github.com/civicrm/civicrm-dev-docs/blob/master/bin/hook-summary.py just regenerates the "All Hooks" summary page

PR #590, opened 2019-03-23
removed reference to the LTS

eileenmcnaughton:
I didn't add in the ESR as I don't believe the policy  is clear enough to be documented but generally I understand that the discussion takes place on the ESR channel & PRs are not really encouraged against the branch - the CT would do a PR if they decide to backport or agree someone else would


mattwire:
This looks good to me.  Be good to get input from core team about whether to add ESR in here but that could be in a follow-on PR.  @colemanw ok to merge?

PR #594, opened 2019-04-03
Buildkit: update php, nodejs, mysql versions

agh1:
https://github.com/civicrm/civicrm-buildkit/pull/441 updates Buildkit to newer npm packages, and it appears that this requires NodeJS 8.x.  This updates the dev docs to reflect that and newer PHP and MySQL versions.

### DIFF
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -75,12 +75,14 @@ You may install buildkit in other environments. The main pre-requisites are:
 
 * Linux or OS X
 * Git
-* PHP 7.0+ (Extensions: `bcmath curl gd gettext imap intl imagick json mbstring mcrypt openssl pdo_mysql phar posix soap zip`)
-* NodeJS (v8 recommended)
+* PHP 5.6+ (7.2+ recommended), including the following extensions: `bcmath curl gd gettext imap intl imagick json mbstring mcrypt openssl pdo_mysql phar posix soap zip`
+* NodeJS 8+
 * NPM
-* Recommended (_for [amp](https://github.com/totten/amp) and [civibuild](/tools/civibuild.md)_)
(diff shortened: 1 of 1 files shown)


docs/tools/buildkit.md
```diff
@@ -75,12 +75,12 @@ You may install buildkit in other environments. The main pre-requisites are:
...
 * Git
-* PHP 5.3+ (Extensions: `bcmath curl gd gettext imap intl imagick json mbstring mcrypt openssl pdo_mysql phar posix soap zip`)
-* NodeJS (v5 recommended)
+* PHP 7.0+ (Extensions: `bcmath curl gd gettext imap intl imagick json mbstring mcrypt openssl pdo_mysql phar posix soap zip`)
+* NodeJS (v8 recommended)
 * NPM
 * Recommended (_for [amp](https://github.com/totten/amp) and [civibuild](/tools/civibuild.md)_)
     * Apache 2.2 or 2.4 (Modules: `mod_rewrite`. On SUSE, possibly `mod_access_compat`. This list may not be exhaustive.)
-    * MySQL 5.1+ (client and server)
+    * MySQL 5.6+ (client and server)
```
totten: @agh1 FWIW, the environments we actively test are [bknix-min](https://github.com/totten/bknix/blob/master/profiles/min/default.nix#L15-L19) (PHP 5.6, MySQL 5.5, NodeJS 8) and [bknix-max](https://github.com/totten/bknix/blob/master/profiles/max/default.nix#L14-L19) (PHP 7.2, MySQL 5.7, NodeJS 8).

There's obviously discussion about mainline branches going up to PHP 7 as a requirement, but even when that happens, there ought be lag between updating `civicrm-core` reqs and `buildkit` reqs. (Folks continue working on backports+extensions even after mainline changes.)

I'd keep the minimums in here aligned with bknix-min (PHP 5.6, MySQL 5.5, NodeJS 8), but the recommendations can be more aggressive.

agh1:
@totten I reworked the format a bit so that we can state the minimum and recommended versions separately.

PR #602, opened 2019-04-15
Add CaseContact to hook_civicrm_post() and hook_civicrm_pre() documen…

aydun:
…tation - see https://github.com/civicrm/civicrm-core/pull/14030

Reposition 'Campaign' in list to be alphabetical


mattwire:
Thankyou @aydun

PR #615, opened 2019-06-03
update intro text

joemcl:


docs/tools/issue-tracking.md
```diff
@@ -1,6 +1,6 @@
 # Issue-tracking systems
 
-As of 2018 CiviCRM uses three different system for tracking and managing issues (aka tickets). This page offers a brief summary of the systems and helps developers understand when and how to use them.
+As of 2018, CiviCRM uses three different systems for tracking and managing issues (aka tickets) - GitLab, GitHub and Jira. This page offers a brief summary of the systems and helps developers understand when and how to use them.
```
totten: At this point, I'm pretty sure we don't use JIRA for anything. (@bgm - isn't `issues.civicrm.org` completely archived now?)

It's probably still good to give a nod to JIRA because it contains historical information, but IMHO it's more correct to say there are only two systems for managing issues.

joemcl: OK @totten yeah seeing this at `issues.civicrm.org` - ```We have migrated from JIRA to Gitlab! New issues must be created on Gitlab at https://lab.civicrm.org/groups/dev/-/issues. For more information: Gitlab roadmap and updates on the migration status. Issue creation was disabled on March 31st 2018. The site was converted to a static archive on December 27 2018.``` So will update.

totten: Nice @joemcl. Looks clearer now.

joemcl: thanks @totten

PR #617, opened 2019-06-06
add text re curl installation to Ubuntu

joemcl:


seamuslee001:
Change seems fine to me

PR #619, opened 2019-06-10
Testing: Update PHPUnit versions

agh1:
Following @totten's PSA on [civicrm-dev](https://lists.civicrm.org/lists/info/civicrm-dev), I figured it would be worth including much of that in the dev docs.  I also updated the examples to use `phpunit5` instead of `phpunit4`.


eileenmcnaughton:
@seamuslee001

PR #623, opened 2019-06-19
Add Docs on new theme hooks

seamuslee001:
While we merged the themeing stub there was no docs on the actual hooks associated this updates it

### DIFF
--- a/docs/hooks/hook_civicrm_activeThemes.md
+++ b/docs/hooks/hook_civicrm_activeThemes.md
@@ -1,0 +1,39 @@
+# hook_civicrm_activeTheme
+
+## Summary
+
+The activeTheme hook determines which theme is active.
+
+## Definition
+
+    hook_civicrm_activeTheme( &$theme, $context)
+
+## Parameters
+
+   - parameter string $theme
+      The identifier for the theme. Alterable.
+      Ex: 'greenwich'.
+   - parameter array $context
+      Information about the current page-request. Includes some mix of:
+      - page: the relative path of the current Civi page (Ex: 'civicrm/dashboard').
+      - themes: an instance of the Civi\Core\Themes service.
+
+## Returns
+
+-   null
+
+## Availability
+
+-   This hook was first available in CiviCRM 5.16
+
+## Example
+
+```php
+     /*
+     * Set which theme is active. 
+     */
+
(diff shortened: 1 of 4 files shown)


totten:
Good idea.

The substance here is copied form `CRM_Utils_Hook`? But made more readable/Markdown-friendly?

seamuslee001:
@totten broadly speaking yes tho i would like some checking on the example code etc

docs/hooks/hook_civicrm_activeThemes.md
```diff
@@ -0,0 +1,39 @@
...
+-   This hook was first available in CiviCRM 5.16
+
+## Example
+
+```php
+     /*
+     * Set which theme is active. 
+     */
+
+    function civitest_civicrm_activeTheme( &$theme, $context ) {
```
totten: Nitpick: does the whitespace and comment formatting deviate from the code-style standards?

seamuslee001: @totten you could make that point about all the hook docs i think

totten: Ah, OK. Agree that should be a separate issue to clean all the hook docs.

docs/hooks/hook_civicrm_themes.md
```diff
@@ -0,0 +1,63 @@
...
+      $themes[] = [
+        'name' => 'civielection',
+        'title' => 'civielection theme',
+        'prefix' => NULL,
+        'url_callback' => '\\Civi\\Core\\Themes\\Resolvers::simple',
+        'search_order' => [
+          0 => 'civielection',
+          1 => '_fallback_',
+        ],
+        'excludes' => [],
```
totten: A couple reactions:

1. The example looks right/valid.
2. The example feel a bit intimidating because it includes all the fields. I'd suggest doing two examples - first, a minimal one; then, a full/complex one.
3. Instead of the string `_fallback_`, use the constant `Civi\Core\Themes::FALLBACK`

seamuslee001: ok @totten updated

totten: Thanks, looks good. 👍

PR #627, opened 2019-06-25
Corrected URL for "unassigned review requests"

twomice:
This URL changed today as we moved these issues to their own gitlab project.


seamuslee001:
looks fine tested the links

PR #631, opened 2019-06-29
Add clarification of what forward-compatibility means

demeritcowboy:
FYI @twomice

### DIFF
--- a/docs/extensions/info-xml.md
+++ b/docs/extensions/info-xml.md
@@ -42,7 +42,7 @@ Here is an example of a full `info.xml` file from [CiviVolunteer](https://github
 
 | CiviCRM Version | Description |
 | -- | -- |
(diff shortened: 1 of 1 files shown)


twomice:
I like this, assuming it's correct; FWIW, it does match my understanding of things. @totten could probably confirm that.

totten:
Yup, @demeritcowboy @twomice , I do believe it's correct. 👍 Good clarification.

One nitpick on wording (emphasis added):

> If you want to specify multiple version compatibility for both lower than 4.7 and higher, then **even if running on an installation of 5.0.0 or higher** you need to specify multiple `<ver>` tags for all of the lower versions and at least one of 4.7 or higher

When I read this, I thought, "oh, run on which installation for whom?". Maybe there's a way to speak about the compatibility without reference to an installation? Ex:

> If your extension has backward compatibility in the 4.x series (Ex: *"from v4.5 through current 5.x"*), then you should specify `<ver>` tags for each previous variant of `4.x`. (Ex: `<ver>4.5</ver><ver>4.6</ver><ver>4.7</ver>`).

demeritcowboy:
I think I wrote the second part before the first part but then yeah now the "installation" part doesn't need to be there. I've updated it. Just trying to figure out how to squash the commits without pulling the entire repo locally, but doesn't seem possible, but not ready to give up on that yet...

twomice:
Looks good to me. @totten okay to merge?

totten:
Agree, looks good

PR #633, opened 2019-07-02
Fix small typos

stevekessler:
I just made some small changes for typos.


totten:
Looks pretty straight-forward. :)

Confirmed the old code rendered wonkily in both mkdocs+Github. New code renders fine in Github.

PR #635, opened 2019-07-05
Add note regarding element inheritence

JKingsnorth:
Certain elements are inherited from the parent 'path' if not defined. This is done in \CRM_Core_Menu::fillMenuValues.


seamuslee001:
This looks good to me merging

PR #637, opened 2019-07-14
hook_civicrm_permission: fix incorrect 'ts' usage.

mlutfy:
Fixes/improves a few things:

* ts() usage in the `$prefix`: punctuation can be localized, ex: in French, there is a non-breaking space before a column.
* In the Civimonitoring example, use `E::ts` instead. I only did it here, since I could clearly show the `use CRM_Civimonitoring_ExtensionUtil as E` without cluttering too much.
* updated the array syntax, because it's not 1995 anymore.

I tried to keep it simple, but the actual "correct" ts usage for this example:

```
// WRONG:
            $prefix = ts('CiviCRM Grant Program:') . ' ';
            $permissions['edit grant programs'] = [
              $prefix . ts('edit grant programs'),                     // label
              ts('Create or edit grant programs and their criteria'),  // description
            ];
```

should be:

```
            $permissions['edit grant programs'] = [
              E::ts('CiviCRM Grant Program: %1', [1 => E::ts('edit grant programs')]),
              E::ts('Create or edit grant programs and their criteria'),
            ];
```

.. but that this would give most people a headache and they just won't use it.

### DIFF
--- a/docs/hooks/hook_civicrm_permission.md
+++ b/docs/hooks/hook_civicrm_permission.md
@@ -29,7 +29,7 @@ to grant it to users.
     -   Simple associative array in the format "permission string =>
         label".  Compatible with CiviCRM 4.3+.
 
-            $prefix = ts('CiviCRM Grant Program') . ': '; // name of extension or module
+            $prefix = ts('CiviCRM Grant Program:') . ' ';
             $permissions['edit grant programs'] = $prefix . ts('edit grant programs');
             $permissions['delete in Grant Program'] = $prefix . ts('delete grant program');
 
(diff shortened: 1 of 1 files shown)


docs/hooks/hook_civicrm_permission.md
```diff
@@ -29,7 +29,7 @@ to grant it to users.
     -   Simple associative array in the format "permission string =>
         label".  Compatible with CiviCRM 4.3+.
 
-            $prefix = ts('CiviCRM Grant Program') . ': '; // name of extension or module
+            $prefix = ts('CiviCRM Grant Program:') . ' ';
```
seamuslee001: @mlutfy just looking at this a couple of thoughts

1. should this be E::ts or similar and
2. the previous code appears to be at least consistant with core https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/Permission.php#L687 not sure if that is correct or not

MikeyMJCO:
What does this need to "unstick" it - anything I can help with?

MikeyMJCO:
@mlutfy @seamuslee001 can we unstick this during the sprint pretty please?

seamuslee001:
Happy to see this merged, have chatted with @mlutfy and will iterate further on this @MikeyMJCO

PR #638, opened 2019-07-15
Add link to more info about the Civi:: facade

demeritcowboy:
Long-term it would be good to add the info contained in the blog post directly into the docs somewhere, but at least in the short term since this is the ONLY mention of Civi::$statics in the dev docs the blog post can at least give more info for now.

@seamuslee001


seamuslee001:
Looks good to me merging

demeritcowboy:
Thanks!

PR #639, opened 2019-07-15
Include documentation on withArray parameter when creating Cache Service

seamuslee001:
This adds a note about the withArray directive when creating a cache service

ping @eileenmcnaughton @totten

### DIFF
--- a/docs/framework/cache.md
+++ b/docs/framework/cache.md
@@ -111,7 +111,12 @@ A few things to notice here:
 * The `name` will be passed down to the storage system -- ensuring that different caches are stored separately.
     * Ex: In `Memcached`/`Redis`, the `name` becomes part of the cache-key.
     * Ex: In `SqlGroup`, the `name` corresponds to the field `civicrm_cache.group_name`.
-* There is an optional `withArray` parameter which if set to `fast` then the cache is wrapped in the `CRM_Utils_Cache_FastArrayDecorator` if type is memory and you have set to use memcache or redis.
+* There is an optional `withArray` parameter with the following acceptable options.
+    * FALSE (default): Reads+writes go directly to the underlying cache
+    * TRUE: There's an extra array-based cache-tier in front of the underlying cache. It uses `CRM_Utils_Cache_ArrayDecorator`.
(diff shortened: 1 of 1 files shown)


totten:
Good idea @seamuslee001 

FWIW, `withArray` accepts three values:

1. `FALSE` (default): Reads+writes go directly to the underlying cache
2. `TRUE`: There's an extra array-based cache-tier in front of the underlying cache. It uses `CRM_Utils_Cache_ArrayDecorator`.
    * This variant is more correct/compliant with PSR-16 in that TTL should be consistent between the front-tier cache and the underlying-cache. It requires a more verbose storage-format, which slightly reduces performance. (*Not measurable for 1-4 reads; would be measureable for 1000 reads.*)
3. `fast`: There's an extra array-based cache-tier in front of the underlying cache. It uses `CRM_Utils_Cache_FastArrayDecorator`.
    * This variant is more performant and uses a cleaner/simpler storage-format; however, you're more likely to get stale reads from the front-tier cache. It's not much of practical drawback in typical usage (where the PHP interpreter only runs for <1s). However, the (in)correctness could be an issue in long-run jobs.

Another small detail - strictly speaking, `withArray` is only used with some drivers:

* `*memory*` drivers (such as Redis, Memcache, APC) respect `withArray`
* `ArrayCache` doesn't use it - because it would be redundant to store the same cache in two in-memory arrays
* `SqlGroup` doesn't use it - because it has its own ad-hoc front-tier cache hard-coded in, and this predates the existence of `withArray`.

eileenmcnaughton:
Also some good notes here - https://github.com/civicrm/civicrm-core/pull/14828 that we should check are adequately incorporated

seamuslee001:
thanks @totten I have included those notes as they are useful so i think this is good to merge now @eileenmcnaughton @MegaphoneJon

MegaphoneJon:
I confirmed that all the notes on civicrm/civicrm-core#14828 are present in the docs already, and Tim's notes are now incorporated, so this looks ready to merge.

eileenmcnaughton:
Great - thanks @seamuslee001 for updating this & @totten for additional input & @MegaphoneJon for merging - yay

PR #641, opened 2019-07-16
Resolve #549 Document Long and short cache

### LINKED ISSUE #549: Document "short" and "long" cache services
See https://github.com/civicrm/civicrm-core/pull/12873.

seamuslee001:
@MegaphoneJon @eileenmcnaughton @totten 

I have included in this PR the commits from #639 as i think it makes more sense to document that and this together but this primarily is about the new paragraph doucmenting the long and short caches. What i don't have here is why someone would use a long v short cache but perhaps that will be fleshed out in review

### DIFF
--- a/docs/framework/cache.md
+++ b/docs/framework/cache.md
@@ -5,13 +5,13 @@
 The easiest way to access a cache is to call `Civi::cache()` and request a built-in cache. There are two useful built-in caches:
 
 * `Civi::cache('long')`: This is for things which are more expensive to refresh. By default, it stores data in a SQL table.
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
I *think* long cache is used for the community news post on the dashboard - ie don't refresh it more than ? once a day because you don't want it fetched for every user who logs in

totten:
@seamuslee001 Agree it's good to document both `withArray` and `long`/`short`.

I think it makes sense to reorder/reframe some of the content in this chapter. At a high-level, the page is currently two sections:

* "Using the default cache" 
* "Using a custom cache"

But now there are two "default" caches, so it's present them differently from the start, e.g.

* "Using a built-in cache"
* "Using a custom cache"

I updated with some draft text to introduce differently and discuss the choice of short-vs-long.

PR #643, opened 2019-07-16
Fix link to test results page

demeritcowboy:


demeritcowboy:
Duplicate of 642 by a few seconds...

Issue #648, opened 2019-07-23
Extension naming convention info is out of date


artfulrobot:
Apparently we no longer use `com.backwards.name` for extensions, preferring simpler names.

So the documentation should be updated.

MikeyMJCO:
I agree :-)

mlutfy:
This was fixed in #649 by @artfulrobot, closing :tada:

PR #649, opened 2019-07-23
Update information on deprecated backwards naming - issue #648

artfulrobot:
This addresses issue #648

### DIFF
--- a/docs/extensions/index.md
+++ b/docs/extensions/index.md
@@ -13,7 +13,7 @@ This section covers how to write extensions.
 
 ## Extension Names
 
-Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names so it should only include characters that can be safely used in that way.
+Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names, so it should only include characters that can be safely used in that way.
 
 ``` info
     Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
--- a/docs/extensions/index.md
+++ b/docs/extensions/index.md
@@ -13,7 +13,9 @@ This section covers how to write extensions.
 
 ## Extension Names
 
-All extension names follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`). For module-extensions, the last word in the module name will be the module's *short-name*. The short-name *must* be unique. It is possible to pick a different short-name, but that requires extra work.
+Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names so it should only include characters that can be safely used in that way.
+
+Nb. Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
 
 ## Pre-Requisites
 
--- a/docs/extensions/index.md
+++ b/docs/extensions/index.md
@@ -15,7 +15,8 @@ This section covers how to write extensions.
 
 Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names so it should only include characters that can be safely used in that way.
 
-Nb. Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
+``` info
+    Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
 
 ## Pre-Requisites
 


docs/extensions/index.md
```diff
@@ -13,7 +13,9 @@ This section covers how to write extensions.
 
 ## Extension Names
 
-All extension names follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`). For module-extensions, the last word in the module name will be the module's *short-name*. The short-name *must* be unique. It is possible to pick a different short-name, but that requires extra work.
+Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names so it should only include characters that can be safely used in that way.
```
MikeyMJCO:
```suggestion
Pick a unique string for your extension's name. Note that this name is used throughout the extension's code, e.g. in function names so it should only include characters that can be safely used in that way.

!!! warning
    There is currently no "canonical list" of extension names. We'd strongly recommend you check the [extensions directory](https://civicrm.org/extensions/) and the [GitLab extensions group](https://lab.civicrm.org/extensions).
```

docs/extensions/index.md
```diff
@@ -13,7 +13,9 @@ This section covers how to write extensions.
...
+
+Nb. Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
```
MikeyMJCO:
```suggestion
``` info
    Historically, extension names used to follow the same convention as Java package names – they look like reversed domain names. (e.g.  `com.example.myextension`), but this is no longer the recommendation.
```

docs/extensions/index.md
artfulrobot: are we allowed underscores in names? I don't see that much, but it would be nice to be clear, e.g. is underscore a "special character" in the above. Also wHaTabOUt case?

I think we're saying it must match `/^[a-z][a-z_0-9]+$/`, really?

MikeyMJCO: Those are good questions @totten @mlutfy can you provide clarity with regards the policy used for GitLab namespaces and in-app limitations?

twomice: Minor readability nitpick: on line 16, could you add a comma beteween "names" and "so"? Then it would read "... code, e.g. in function names, so ..."

artfulrobot: @twomice  not nitpicking at all, Yes, it needs that! I was just waiting on the above to be resolved. And I'm not even sure how you edit things ↑ here like @MikeyMJCO did.

twomice: > how you edit things ↑ here like @MikeyMJCO did.

Yeah, I was wondering the same thing.

MikeyMJCO: > > how you edit things ↑ here like @MikeyMJCO did.
> 
> Yeah, I was wondering the same thing.

In "Review"/"Files Changed" when you over over a line you'll get a + button - you can then click the little "Insert a Suggestion" button (CTRL+G)

MikeyMJCO:
```suggestion
Pick a unique single word for your extension's name. Note that this name is used throughout the extension's code, e.g. function names, so it should only include characters that can be safely used in that way.
```

artfulrobot: Doh! I committed the suggestion and it 'resolved' the whole 'conversation' including the bits we're waiting on some others to input on. I've 'unresolved' it again in case resolving it put the core team off from inputting.

mattwire:
Don't let it hold up this PR but it would be nice to extend with a bit about renaming extensions - as changing the extension key isn't easy to do for an existing extension - and probably not recommended.

michaelmcandrew:
I'd quite like to put together some best practice on how to rename an extension, what the dangers are, think about if and how we can make it less disruptive. There is a wordpress plugin that does view counts that was renamed (I forget the name) that has a bit of logic in to detect previous versions. Might be useful as some inspiration.

totten:
IIRC, there's some ancient and ill-advised logic that allows certain fields to reference "(partial) class-name *or* extension key". This means you're given string and need to figure out if it's a class or ext-key. To wit, ask: "*Is there dot in this string?*"  (`Mapper::isExtensionKey($string)`). My *hunch* is that this check only matters for legacy extension-types (`type="payment"`, `type="report", etc).

Which leads to this (draft/unvalidated) rule of thumb:

* For legacy extension types (`payment`, `report`, etc), it is *mandatory* to have a dot in the name.
* For normal `module` extensions, it is *optional* to have a dot in the name.

If one really wants to put a nail in the coffin of dotted names, then I'd suggest auditing that rule of thumb. Perhaps grep  `isExtensionKey()` and consider how each use-case is addressed by (a) legacy extensions vs (b) module extensions.

```
$ svngrep isExtensionKey .
./web/sites/all/modules/civicrm/Civi/Payment/System.php:      if ($ext->isExtensionKey($processor['class_name'])) {
./web/sites/all/modules/civicrm/Civi/Payment/System.php:    if ($ext->isExtensionKey($processor['class_name'])) {
./web/sites/all/modules/civicrm/CRM/Contact/Selector/Custom.php:    if (!$ext->isExtensionKey($customSearchClass)) {
./web/sites/all/modules/civicrm/CRM/Contact/BAO/SearchCustom.php:    if (!$ext->isExtensionKey($customSearchClass)) {
./web/sites/all/modules/civicrm/CRM/Core/Payment.php:      if ($ext->isExtensionKey($dao->class_name)) {
./web/sites/all/modules/civicrm/CRM/Extension/Mapper.php:  public function isExtensionKey($key) {
./web/sites/all/modules/civicrm/CRM/SMS/Provider.php:      if ($ext->isExtensionKey($providerName)) {
```

Re: infra - It is possible that the extension-directory has some validation-logic that looks for dots, but this will be probably become irrelevant or relaxed after https://lab.civicrm.org/extensions/extensions-directory/issues/38 is resolved.

twomice:
@artfulrobot I agree with the suggestions @MikeyMJCO has made. Any thoughts?

twomice:
Good question. Maybe @MikeyMJCO has some idea about the specific mechanics of the appropriate next steps to get this a) changed per his suggestions and b) merged.

MikeyMJCO:
> 
> 
> @twomice yeah fine, I thought it was all done, but I find github v confusing with these things - it still says "changes requested". Do I have to press ReRequest review or something?

The answer is no - I also find GH confusing with this in that I as the reviewer have to come back and change my review to "approve" the changes. GL is better at this because when the requested changes are implemented it auto resolves the review.

MikeyMJCO:
I'm merging this change as is - will copy out the discussion on renaming extensions to a separate issue. I think @totten's comment could be a standalone issue against core for the longer term work to deprecate dependencies on the long-form naming.

PR #664, opened 2019-08-21
civilint.md - Split out comments on multiple lines

totten:
These examples are in a code block  - which prevents wrapping. When reading on the published site, you have to scroll horizontally.

Moving them to separate lines makes them more legible.


MikeyMJCO:
Much clearer. :+1:

PR #667, opened 2019-08-31
clarification on running curl as sudo

joemcl:
Added to line 22 - "You shouldn't run the curl buildkit command as sudo. It will sudo itself for the bits that need it."

### DIFF
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -19,7 +19,7 @@ curl -Ls https://civicrm.org/get-buildkit.sh | bash -s -- --full --dir ~/buildki
 
 !!! note
 
-    * When executing the above command, you should *not* run as `root`, as it *will* cause failures. However, you *should*     have `sudo` permissions.
+    * When executing the above command, you should *not* run as `root`, as it *will* cause failures. However, you *should*     have `sudo` permissions. You shouldn't run the curl buildkit command as sudo. It will sudo itself for the bits that need it.
     * The `--full` option is *very opinionated*; it specifically installs `php`, `apache`, and `mysql` (rather than `hhvm`, `nginx`, `lighttpd`, or `percona`). If you try to mix `--full` with alternative systems, then expect conflicts.
     * If you use the Ubuntu feature for "encrypted home directories", then don't put buildkit in `~/buildkit`. Consider `/opt/buildkit`, `/srv/buildkit`, or some other location that remains available during reboot.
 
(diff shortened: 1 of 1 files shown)


docs/tools/buildkit.md
```diff
@@ -19,7 +19,7 @@ curl -Ls https://civicrm.org/get-buildkit.sh | bash -s -- --full --dir ~/buildki
 
 !!! note
 
-    * When executing the above command, you should *not* run as `root`, as it *will* cause failures. However, you *should*     have `sudo` permissions.
+    * When executing the above command, you should *not* run as `root`, as it *will* cause failures. However, you *should*     have `sudo` permissions. You shouldn't run the curl buildkit command as sudo. It will sudo itself for the bits that need it.
```
MikeyMJCO:
```warning
    * When executing the above curl/install command, you should *not* run it as `root` - including via `sudo`, as doing so *will* cause failures. However, you *should* have `sudo` permissions as the script includes several `sudo` statements where these are required.
```

totten: Ah, so it sounds like `sudo` has become so ubiquitous as the means of becoming `root` that people may not reflexively equate `root =~ sudo`.

Agree with @MikeyMJCO that it's better to rephrase the note generally (instead of appending).

totten: In the old/original text, I felt awkward about emphasizing *what-not-to-do* (negative tone) and settled for that text mainly because it was short. If we can afford a few more words, then maybe we can flip the tone? Like...

```suggestion
    * The above command installs some resources as your regular user (e.g.`/home/myuser/buildkit`) and others as the `root` user (e.g. the `mysqld` system service).
    * To ensure proper permissions, you *must* run the installer as your *regular user* -- **not** `root`/`sudo`.  If the installer requires elevated permissions, it will automatically call `sudo` and prompt for a password.
```

Or is that too verbose?

totten: (*Random note: I found it hard to read the suggestion blocks in Github's "Conversation" tab - because the text isn't wrapped. But if you switch to the "Files changed" tab, they become easier.*)

joemcl: thanks, works for me!

joemcl:
hey @totten @MikeyMJCO not sure if I've screwed this up or not - my intent is to accept the changes from both of you...

MikeyMJCO:
It's great Joe! Merging!

joemcl:
thanks @MikeyMJCO

PR #674, opened 2019-09-24
Add Queue Reference

artfulrobot:
I found it hard to figure out how queues work as it was not included in the docs. Here is a single page reference which includes

- basic concepts
- links to the demoqueue extension (which I wish I'd found sooner!)
- cautions for programmers (based on painful experience!)

PR #676, opened 2019-09-24
Queue section - Misc edits

totten:
1. Cleaning up list of definitions
2. Update `create()` docs, reframing `type` and `reset` in a more-matter-of-fact/less-surprised way.

### DIFF
--- a/docs/framework/queues/index.md
+++ b/docs/framework/queues/index.md
@@ -14,6 +14,7 @@ extension that shows how to use an SQL-based queue.
 ## Definitions
 
 - *Queue*: An object representing a queue, which may be new or existing.
+- *Queue type*: A storage backend for the list of queue items.
 - *Queue runner*: A class for processing the queued items.
 - *Item*: A single job on a queue
 - *Task*: A particular type of item, as expected by CiviCRM's queue runner.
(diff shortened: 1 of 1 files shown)


totten:
This is a follow-up to #674; cc @artfulrobot

MikeyMJCO:
@artfulrobot if you approve the changes I'll go ahead and merge this.

totten:
@artfulrobot That's a big question. I think it'll be just as big in any medium. 🙃 

I've been re-reading the queue code from that POV, and I think there are a couple things to consider:

* __Use-Cases/Main-Loops__: Much of the queue system was originally added to support the upgrader. In that use-case, the goal is to use many small PHP requests/tasks to mitigate the risk/damage of HTTP timeouts. You build-up a list of tasks, execute them all sequentially, and then it's done. A different use-case would be background email notifications - whenever something happens on the frontend ("New contribution!"), you enqueue a background task to send an email (which can be slow; depending on network I/O). This task list is very open-ended; there can be multiple producers+consumers; and it's never really done or closed. The main-loops for these would be different -- e.g. `runAll()` works for the first; `civicrm_api_job_runspecificqueue()` works for the second; and both would be broken if you tried to swap them.
    * __Specifically__: If you tried use `runAll()` in a use-case that was more open-ended, then I think you could see some raciness around `numberOfItems`/`is_completed`. But if you choose a more suitable main-loop, then these are irrelevant.
    * __Aside__: The core code provides a main-loop for the first scenario (`runAll()`), but it doesn't provide a main-loop for the second scenario. I really like that `civicrm_api_job_runspecificqueue()` shows a runner/main-loop for the latter use-case. It'd be tempting to have a couple other runners/mainloops available in core.
* __SQL Implementation, Transactions, Locks__: Some thoughts:
    * For the upgrader/run-all use-cases, I think it's less demanding about correct usage of transactions+locks. (I've never seen a report of upgrades failing due to deadlocks in the task queue...)
    * For the email/background-notification use-cases, the *contract* of `CRM_Queue_Queue_Sql` is probably fine, but the *implementation* looks a bit off. There are only a handful of SQL queries in there, but I suspect some need to be rewritten and thoughtfully tested.
    * When a consumer calls `claimItem()`, it should *not* be inside an active/long-running transaction. (AFAIK, this is currently OK.) 
    * When a consumer calls `claimItem()`, it should be atomic, so it probably needs to use a brief transaction.
    * When a producer calls `createItem()`, I think it's preferable to be able to use that inside or outside of transactions. (In the example of sending email notifications about contributions, you might call `createItem(...sendThankYouEmail...)` as part of the transaction. But if the commit subsequently rolls back, then it doesn't make sense to send a thank you email.) This is a nice-to-have, imho, and there are fallback options... if experience shows that you shouldn't use `createItem()` within a transaction, then move it to a [post-commit](https://github.com/civicrm/civicrm-core/pull/15338) phase.

totten:
Right, some kind of locking-read sounds appropriate. For a first guess, I would'a gone for two changes...

1. Use `SELECT ... FOR UPDATE ...` (because there is an update...)
2. Include an extra `where` based on `release_time`. `claimItem()` should only take things that are ripe for usage.

However, I wouldn't really commit to that. Instead, one would budget their time to (a) figure out a proper test procedure and (b) twiddle the queries for a bit.

NB to run several concurrent workers on top of  `civicrm_queue_item`, one would have locking on two-levels. On a high-level, the `civicrm_queue_item.release_time` allows each worker to hold onto a job for seconds/minutes/hours/whatever. On a low-level, for a few milliseconds, it needs that InnoDB locking to control races among active threads.

totten:
@artfulrobot Brilliant! I downloaded it and tried it on `Sql` and `Sql2`. :)

We should probably add that driver to core, yeah? More thoughts:

1. Using a new class name (like `Sql2`) is good. At first, I sorta thought, "Why would you want *two* SQL-backed drivers?" But they both have use-cases; the old query makes more sense if you want tasks to be strictly sequential (like in the upgrader UI - where consistent ordering is important), and the new query makes sense if you want tasks to be executed in parallel (like in an email-notification use-case). Maybe we could find a name which evokes this multithreaded usage (e.g. `ParaSql` or `ConcurrentSql` or `MinimallyBlockingSql`)?

2. What should we do with the test long-term? This is hugely valuable for maintainability. It's technically a bit different from the others. (I can relate, of course.... bash has more job-control features built-in...) 

    * We could import the current script and revise it to provide CI outputs. The main thing is to have semantic exit-codes (`exit 0`=>OK; `exit 1`=>Bad) and to generate a JUnit XML file. (Ex: [bash+junit](https://github.com/civicrm/civicrm-upgrade-test/blob/master/civicrm-upgrade-test#L229-L265))
    * It is possible to do multithreaded testing in PHPUnit (ex: [api_v3_JobProcessMailingTest::testConcurrency](https://github.com/civicrm/civicrm-core/blob/5.18/tests/phpunit/api/v3/JobProcessMailingTest.php#L367-L424) uses `createExternalAPI()` aka [Civi\API\ExternalBatch](https://github.com/civicrm/civicrm-core/blob/5.18/Civi/API/ExternalBatch.php)), and it's nice to use the same test-runner, but that's a re-write and we might lose something in the process.
    * We could keep the test as-is in a separate repo, and sprinkle some comments into `CRM/Queue/*`  and [review#gotchas](https://docs.civicrm.org/dev/en/latest/standards/review/#gotchas). This may actually be the best cost/benefit...

PR #677, opened 2019-09-29
Add token/tokenValues hook examples

mattwire:
`tokenValues` gets slightly different parameters when called from different places.  That's probably a bug but for now we should actually add an example to the docs to show how it needs to be implemented.


mattwire:
@MikeyMJCO Thanks I've made the requested changes and corrected some other bits

MikeyMJCO:
That's great - thanks Matt :-)

PR #686, opened 2019-10-08
Further improvements to l10n format in hook_civicrm_permission docume…

seamuslee001:
…ntation

This adds some additional fixes and more standardises what should be expected even tho core may not be adopting these patterns right now. Also encases the code in php syntax highlighting


MikeyMJCO:
I'm merging this as-is - if it need further tweaking this can be done in further PR(s).

Issue #520, opened 2018-04-10
Make extension review process compatible with GitLab


seancolsen:
1. The extension review process described here...

    * https://docs.civicrm.org/dev/en/latest/extensions/publish/#automated-distribution
    * https://docs.civicrm.org/dev/en/latest/extensions/lifecycle/#formal-review

    ...is dependent on Jira. 

1. Jira reads:

    > Issue creation was disabled on March 31st 2018.

These points seem incompatible to me. 

Can we offer an extension review process that doesn't depend on Jira? 

@GinkgoFJG @twomice any thoughts?

twomice:
Good question. I'll raise this with Frank at our next call.

Thanks,
Allen

On Tue, 10 Apr 2018 18:36:14 +0000 (UTC)
Sean Madsen <notifications@github.com> wrote:

> 1. The extension review process described here...
> 
>     *
> https://docs.civicrm.org/dev/en/latest/extensions/publish/#automated-distribution
>     *
> https://docs.civicrm.org/dev/en/latest/extensions/lifecycle/#formal-review
> 
>     ...is dependent on Jira. 
> 
> 1. Jira reads:
> 
>     > Issue creation was disabled on March 31st 2018.
> 
> These points seem incompatible to me. 
> 
> Can we offer an extension review process that don't depend on Jira? 
> 
> @GinkgoFJG @twomice any thoughts? 
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/civicrm/civicrm-dev-docs/issues/520


-- 
Joinery: Mission-critical data solutions
Migrate, integrate, report.

Allen Shaw <allen@joineryhq.com>
903-361-7429

joemcl:
Related to #606 . Anything happening with this @twomice ?

joemcl:
So this is 'fixed' with this right? - https://lab.civicrm.org/extensions/extension-review-requests/issues ?

twomice:
@joemcl There's still one reference to JIRA in https://docs.civicrm.org/dev/en/latest/extensions/lifecycle , for which I've submitted PR #687

twomice:
PR #687 is merged now (thanks @MikeyMJCO!). So this is fixed. Apparently I don't have rights to close.

Request to close this ticket @joemcl

joemcl:
thanks @MikeyMJCO and thanks for closing @seamuslee001

PR #687, opened 2019-10-08
Removing references to JIRA, per #520

twomice:
RE: https://github.com/civicrm/civicrm-dev-docs/issues/520


MikeyMJCO:
This is a no-brainer - merging. Thanks @twomice

PR #700, opened 2019-10-16
Add in Examples and changes chapters for APIv4

seamuslee001:


seamuslee001:
@MikeyMJCO your changes were made also ping @colemanw

colemanw:
I'd close this in favor of #708

colemanw:
@seamuslee001 I appreciate the thoroughness but considering the low (miniscule) adoption rate by developers I think we can consider the extension to be an unstable "alpha" and not document changes made prior to merging into core. Otherwise this list would be veeeery long if we went through every change made to the extension over the past 2 years.

PR #704, opened 2019-10-19
Copy editing on Order api

eileenmcnaughton:
I just read the first part & believe this edit makes it more succinct & clearer

@JoeMurray @artfulrobot

### DIFF
--- a/docs/financial/OrderAPI.md
+++ b/docs/financial/OrderAPI.md
@@ -3,7 +3,7 @@
     will be documented and the core code underlying this area may change from version
     to version.
 
-An 'order' is our developer term for a pseudo-entity that maps to the CiviCRM contribution object but also encompasses related entities like memberships, event registrations and unlying financial entities. There is no single 'order' table but the top level order information is stored in the civicrm_contribution table.
+An 'order' is our developer term for a pseudo-entity that maps to the CiviCRM contribution object but also encompasses related entities like memberships, event registrations and underlying financial entities. There is no single 'order' table but the top level order information is stored in the civicrm_contribution table.
 
 The Order API is intended to be used as the primary API for adding, updating, and deleting orders. When using the order api you should
 
--- a/docs/financial/OrderAPI.md
+++ b/docs/financial/OrderAPI.md
@@ -3,21 +3,14 @@
     will be documented and the core code underlying this area may change from version
     to version.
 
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
" the order api is also to be used for updates"

Honestly - that is news to me - I was never under the  impression the requirements for that had been finalised so I removed it  -  unless you can confirm it IS tested & reliable there.

eileenmcnaughton:
@artfulrobot I made some  further tweaks - - https://github.com/civicrm/civicrm-dev-docs/pull/704/files#diff-578bec3cae0ba7a26b7048272e5e5055- but I think we should remove reference to  using Order api to update until we know it does that reliably -  it's not tested and not used for that in any core code and I don't think we should promote that it does something that I don't think we can say it reliably does or that the code has really been writtten for.

I would like to see us mature Order.update in apiv4 but documenting the create flow feels like our current priority

docs/financial/OrderAPI.md
```diff
@@ -3,21 +3,20 @@
     will be documented and the core code underlying this area may change from version
     to version.
 
-The Order API is intended to be used as the primary API for adding, updating, and deleting orders.
+An 'order' is our developer term for a pseudo-entity that maps to the CiviCRM contribution object but also encompasses related entities like line items, memberships, event registrations and underlying financial entities. There is no single 'order' table but the top level order information is stored in the civicrm_contribution table.
 
-An 'order' is a non-CiviCRM term that corresponds to how CiviCRM uses its contribution object in terms of handling the full life-cycle of a purchase of memberships, event registrations or making a donation. Unlike most APIs, there is no table directly associated with the Order API.
+The Order API is intended to be used as the primary API for adding, updating, and deleting orders. When using the order api you should
```
MikeyMJCO:
```suggestion
The Order API is intended to be used as the primary API for adding, updating, and deleting orders. When using the Order API you should:
```

MikeyMJCO: Dangit - forgot semi colon after "should not" as well.

eileenmcnaughton:
@MikeyMJCO I think this is OK now? I think we should use more of an edit-wars approach to the financial docs while they are still quite fluid & then get more strict on review after - I think I have 3 PRs open on them

MikeyMJCO:
Agreed - I'm 100% in favour of a "living docs" approach :-)

MikeyMJCO:
Meging!

PR #706, opened 2019-10-20
Add in documentation on APIv4 Options

seamuslee001:


docs/api/v4/options.md
```diff
@@ -0,0 +1,86 @@
+# APIv4 Options
+
+There are many API Options accepted by the CiviCRM API. These options allow the developer to add in more parameters to the resulting Query that is run against the database. E.g. Limit, Sort. You can explore these options using the the [API Explorer](/api/index.md#api-explorer) Unlike in APIv3 setting the options is not contained within an array of options in the parameters array.
```
MikeyMJCO: This is worded a little unclearly towards the end.

docs/api/v4/options.md
```diff
@@ -0,0 +1,86 @@
...
+
+
+## Orderby
+
+-   **Action**: get
+-   **Type**: string
+-   **Default** id ASC
+
+Determine what the orer of the records should be. This was either done with `options.sort` or `options.sequential` in APIv3
```
MikeyMJCO:
```suggestion
Determine what the order of the records should be. This was either done with `options.sort` or `options.sequential` in APIv3
```

docs/api/v4/options.md
```diff
@@ -0,0 +1,86 @@
+# APIv4 Options
+
+There are many API Options accepted by the CiviCRM API. These options allow the developer to add in more parameters to the resulting Query that is run against the database. E.g. Limit, Sort. You can explore these options using the the [API Explorer](/api/index.md#api-explorer) Unlike in APIv3 setting the options is not contained within an array of options in the parameters array.
+
+
+## Orderby
```
MikeyMJCO:
```suggestion
## OrderBy
```

seamuslee001:
@MikeyMJCO added in your suggestions and have re-written the last sentence of the opening paragraph to hopefully be a bit better but appreaciate your thoughts

docs/api/v4/options.md
```diff
@@ -0,0 +1,86 @@
+# APIv4 Options
+
+There are many API Options accepted by the CiviCRM API. These options allow the developer to add in more parameters to the resulting Query that is run against the database. E.g. Limit, Sort. You can explore these options using the the [API Explorer](/api/index.md#api-explorer) In APIv3 options were set as keys of an array called options within the parameters array. In APIv4 tthey are just passed in as another parameter not in a nested array.
```
totten: Some copy-edits:

```suggestion
The CiviCRM API accepts many *options*. These options allow the developer to refine the behavior of the API call -- for example, setting a *limit* on the number of records or the *order* of the returned records.

Every API may support distinct options. In this page, we'll examine a few of the most common options in APIv4. For the most detailed information about available options, consult the [APIv4 Explorer](/api/index.md#api-explorer).
```

Substantively, this suggestion does strike the comparison between APIv3/APIv4. The comparison is a good idea, but I'm wondering if maybe it should be a separate document. For someone reading *today*, the comparison is probably more important than the reference-material for APIv4. OTOH, if someone is reading this in, say, 12 months -- then the comparisons would be by-the-by.

Maybe we need a document like "APIv4 for People Who Know APIv3". (Very rough cut: https://gist.github.com/totten/cd2e62e32c84fa6e1b0a1021ffe0a252) @seamuslee001 @colemanw @MikeyMJCO - I wonder if that kind of content belongs in the Dev Guide - or perhaps maybe more appropriate to a blog?

MikeyMJCO: The suggested edit is good. I think the content you describe would be more appropriate as a blog. We can link to it from the dev guide if appropriate.

colemanw:
I don't think "Options" is the right name for this section. As Tim [noted in his gist](https://gist.github.com/totten/cd2e62e32c84fa6e1b0a1021ffe0a252), the whole "options" thing was kind of shoehorned into api3. Api4 has a smarter design so we don't need a param named "options" we just have params.

Now that I read through the content, I'm not sure we need this section at all. The big problem with api3 options (well, not the only problem, but one of them) was that they were not self-documenting and were not very discoverable, so we needed a page like this to document them. But that's not the case in v4.

Honestly I don't think this page is needed.

colemanw:
I'd like to close this in favor of #708

PR #708, opened 2019-10-21
Add information about api4 changes

colemanw:
This adds a changelog and information adapted from https://github.com/civicrm/org.civicrm.api4/blob/master/readme.md and https://gist.github.com/totten/cd2e62e32c84fa6e1b0a1021ffe0a252

### DIFF
--- a/docs/api/v4/differences-with-v3.md
+++ b/docs/api/v4/differences-with-v3.md
@@ -1,31 +1,32 @@
-# Differences between Api v3 and v4
+# Differences between API v3 and v4
 
-If you are already familiar with Api v3, here are the key differences to help you get started using v4.
+If you are already familiar with APIv3, here are the key differences to help you get started using APIv4.
 
-## Api wrapper
-- In addition to the familiar style of `civicrm_api4('Entity', 'action', $params)` there is now an OO style in php `\Civi\Api4\Entity::action()->...->execute()`.
-- When chaining api calls together, backreferences to values from the main api call must be explicitly given (discoverable in the api explorer).
-- `$checkPermissions` always defaults to `TRUE`. In api3 it defaulted to `TRUE` in REST/Javascript but `FALSE` in php.
+## API wrapper
+- In addition to the familiar style of `civicrm_api4('Entity', 'action', $params)`, there is now an object-oriented style in PHP: `\Civi\Api4\Entity::action()->...->execute()`
+- When chaining API calls together, back-references to values from the main API call must be explicitly given (discoverable in the API Explorer).
+- `$checkPermissions` always defaults to `TRUE`. In APIv3, the default depended on the environment (`TRUE` in REST/Javascript; `FALSE` in PHP).
 - A 4th param `index` controls how results are returned:
-  Passing a string will index all results by that key e.g. `civicrm_api4('Contact', 'get', $params, 'id')` will index by id.
-  Passing a number will return the result at that index e.g. `civicrm_api4('Contact', 'get', $params, 0)` will return the first result and is the same as `\Civi\Api4\Contact::get()->execute()->first()`. `-1` is the equivalent of `$result->last()`.
+    - Passing a string will index all results by that key e.g. `civicrm_api4('Contact', 'get', $params, 'id')` will index by id.
+    - Passing a number will return the result at that index e.g. `civicrm_api4('Contact', 'get', $params, 0)` will return the first result and is the same as `\Civi\Api4\Contact::get()->execute()->first()`. `-1` is the equivalent of `$result->last()`.
 
 ## Actions 
-- `Get` no longer sets a default limit of 25 outside the api explorer.
-- The `Create` action is now only used for creating new items (no more implicit update by passing an id to v3 `create`).
+- For `Get`, the default `limit` has changed. If you send an API call without an explicit limit, then it will return *all* records. (In v3, it would silently apply a default of 25.) However, if you use the API Explorer, it will *recommend* a default limit of 25.
+- The `Create` action is now only used for creating *new* items (no more implicit update by passing an id to v3 `create`).
 - The `Save` action in v4 is most similar to v3's `create` - it accepts one or more records to create or update, infering the action based on the presence of `id` in each record.
 - `Update` and `Delete` can be performed on multiple items at once by specifying a `where` clause, vs a single item by id in v3.
 - `getsingle` is gone, use `$result->first()` or `index` `0`.
 - `getoptions` is no longer a standalone action, but part of `getFields`.
 
 ## Output  
-- Output is an array with object properties rather than a nested array.
-- In PHP, you can `foreach` the results arrayObject directly, or you can call methods on it like `$result->first()` or `$result->indexBy('foo')`.
-- By default, results are indexed sequentially like api3 `sequential => 1`, but the `index` param or `indexBy()` method let you change that.
-  e.g. `civicrm_api4('Contact', 'get', $params, 'id')` or `\Civi\Api4\Contact::get()->execute()->indexBy('id')` will index results by id.
+- Output is an object (`Result` aka [`ArrayObject`](https://www.php.net/manual/en/class.arrayobject.php)) rather than a plain `array`.
+- In PHP, you can iterate over the `ArrayObject` (`foreach ($myResult as $record)`), or you can call methods like `$result->first()` or `$result->indexBy('foo')`.
+- By default, results are indexed sequentially (`0,1,2,3,...` - like APIv3's `sequential => 1`). You may optionally index by `id`, `name`, or any other field, as in:
+    - (Procedural-style; use `$index` parameter): `civicrm_api4('Contact', 'get', [], 'id')`
+    - (OOP-style; use `indexBy()` method): `\Civi\Api4\Contact::get()->execute()->indexBy('id')`
 
 ## Input
-- Instead of a single `$params` array containing a mishmash of fields, options and parameters, each api4 parameters is distinct (see section on Params below).
+- Instead of a single `$params` array containing a mishmash of fields, options and parameters, each APIv4 parameter is distinct (see section on Params below).
 - Custom fields are refered to by name rather than id. E.g. use `constituent_information.Most_Important_Issue` instead of `custom_4`.
 
 ## Params
--- a/docs/api/v4/differences-with-v3.md
+++ b/docs/api/v4/differences-with-v3.md
@@ -1,4 +1,4 @@
-# Differences between API v3 and v4
+# Differences Between API v3 and v4
 
 APIv4 is broadly similar to APIv3. Both are designed for reading and writing data.
 Both use *entities*, *actions*, and *parameters*. However, APIv4 is specifically a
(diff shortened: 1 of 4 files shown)


colemanw:
TODO: I think we should move the Hooks section out of the Api3 changelog and into its own page, probably under Hooks.

colemanw:
Ping @totten @seamuslee001

totten:
I pushed up some copy-edits.

Also, I think it's helpful to start with concrete examples, so I visual up a table at the top. But maybe this means that my section toward the bottom can be trimmed-down now

seamuslee001:
Looks all fine to me thoughts @MikeyMJCO

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+
+APIv4 is broadly similar to APIv3. Both are designed for reading and writing data.
+Both use *entities*, *actions*, and *parameters*. However, APIv4 is specifically a
+breaking-change which aims to reduce *ambiguity* and improve *flexibility* and *consistency*.
+
+This document walks through a list of specific differences.  As you consider
+them, it may help to have a concrete example expressed in both APIv3 and
+APIv4:
+
+<!-- Would be nice if Markdown made it easier to do side-by-side comparison... -->
```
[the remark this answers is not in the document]
MikeyMJCO: With the above said - let's get it in and I can fix the table afterwards - not critical!

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+12:   echo $row['display_name'];
+13: }
+</pre></div>
+</td>
+    </tr>
+  </tbody>
+</table>
+
+
+## API wrapper
```
MikeyMJCO:
```suggestion
## API Wrapper
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+
+- APIv4 supports two notations in PHP:
+    - Procedural/array style: `civicrm_api4('Entity', 'action', $params)`
```
MikeyMJCO:
```suggestion
    * Procedural/array style: `civicrm_api4('Entity', 'action', $params)`
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - Object-oriented style: `\Civi\Api4\Entity::action()->...->execute()`
```
MikeyMJCO:
```suggestion
    * Object-oriented style: `\Civi\Api4\Entity::action()->...->execute()`
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- When using OOP style in an IDE, most actions and parameters can benefit from auto-completion and type-checking.
```
MikeyMJCO:
```suggestion
* When using OOP style in an IDE, most actions and parameters can benefit from auto-completion and type-checking.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+</pre></div>
+</td>
+    </tr>
+  </tbody>
+</table>
+
+
+## API wrapper
+
+- APIv4 supports two notations in PHP:
```
MikeyMJCO:
```suggestion
* APIv4 supports two notations in PHP:
```

Docs use * for bullets not - according to style guide.

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - Procedural/array style: `civicrm_api4('Entity', 'action', $params)`
+    - Object-oriented style: `\Civi\Api4\Entity::action()->...->execute()`
+- When using OOP style in an IDE, most actions and parameters can benefit from auto-completion and type-checking.
+- `$checkPermissions` always defaults to `TRUE`. In APIv3, the default depended on the environment (`TRUE` in REST/Javascript; `FALSE` in PHP).
```
MikeyMJCO:
```suggestion
* `$checkPermissions` always defaults to `TRUE`. In APIv3, the default depended on the environment (`TRUE` in REST/Javascript; `FALSE` in PHP).
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- A 4th param `index` controls how results are returned:
```
MikeyMJCO:
```suggestion
* A 4th param `index` controls how results are returned:
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - Passing a string will index all results by that key e.g. `civicrm_api4('Contact', 'get', $params, 'id')` will index by id.
```
MikeyMJCO:
```suggestion
    * Passing a string will index all results by that key e.g. `civicrm_api4('Contact', 'get', $params, 'id')` will index by id.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - Passing a number will return the result at that index e.g. `civicrm_api4('Contact', 'get', $params, 0)` will return the first result and is the same as `\Civi\Api4\Contact::get()->execute()->first()`. `-1` is the equivalent of `$result->last()`.
```
MikeyMJCO:
```suggestion
    * Passing a number will return the result at that index e.g. `civicrm_api4('Contact', 'get', $params, 0)` will return the first result and is the same as `\Civi\Api4\Contact::get()->execute()->first()`. `-1` is the equivalent of `$result->last()`.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- When chaining API calls together, back-references to values from the main API call must be explicitly given (discoverable in the API Explorer).
```
MikeyMJCO:
```suggestion
* When chaining API calls together, back-references to values from the main API call must be explicitly given (discoverable in the API Explorer).
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Actions 
+- For `Get`, the default `limit` has changed. If you send an API call without an explicit limit, then it will return *all* records. (In v3, it would silently apply a default of 25.) However, if you use the API Explorer, it will *recommend* a default limit of 25.
```
MikeyMJCO:
```suggestion
* For `Get`, the default `limit` has changed. If you send an API call without an explicit limit, then it will return *all* records. (In v3, it would silently apply a default of 25.) However, if you use the API Explorer, it will *recommend* a default limit of 25.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- The `Create` action is now only used for creating *new* items (no more implicit update by passing an id to v3 `create`).
```
MikeyMJCO:
```suggestion
* The `Create` action is now only used for creating *new* items (no more implicit update by passing an id to v3 `create`).
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- The `Save` action in v4 is most similar to v3's `create` - it accepts one or more records to create or update, infering the action based on the presence of `id` in each record.
```
MikeyMJCO:
```suggestion
* The `Save` action in v4 is most similar to v3's `create` - it accepts one or more records to create or update, infering the action based on the presence of `id` in each record.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- `Update` and `Delete` can be performed on multiple items at once by specifying a `where` clause, vs a single item by id in v3.
```
MikeyMJCO:
```suggestion
* `Update` and `Delete` can be performed on multiple items at once by specifying a `where` clause, vs a single item by id in v3.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- `getsingle` is gone, use `$result->first()` or `index` `0`.
```
MikeyMJCO:
```suggestion
* `getsingle` is gone, use `$result->first()` or `index` `0`.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- `getoptions` is no longer a standalone action, but part of `getFields`.
```
MikeyMJCO:
```suggestion
* `getoptions` is no longer a standalone action, but part of `getFields`.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Output  
+- Output is an object (`Result` aka [`ArrayObject`](https://www.php.net/manual/en/class.arrayobject.php)) rather than a plain `array`.
```
MikeyMJCO:
```suggestion
* Output is an object (`Result` aka [`ArrayObject`](https://www.php.net/manual/en/class.arrayobject.php)) rather than a plain `array`.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- In PHP, you can iterate over the `ArrayObject` (`foreach ($myResult as $record)`), or you can call methods like `$result->first()` or `$result->indexBy('foo')`.
```
MikeyMJCO:
```suggestion
* In PHP, you can iterate over the `ArrayObject` (`foreach ($myResult as $record)`), or you can call methods like `$result->first()` or `$result->indexBy('foo')`.
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- By default, results are indexed sequentially (`0,1,2,3,...` - like APIv3's `sequential => 1`). You may optionally index by `id`, `name`, or any other field, as in:
```
MikeyMJCO:
```suggestion
* By default, results are indexed sequentially (`0,1,2,3,...` - like APIv3's `sequential => 1`). You may optionally index by `id`, `name`, or any other field, as in:
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - (Procedural-style; use `$index` parameter): `civicrm_api4('Contact', 'get', [], 'id')`
```
MikeyMJCO:
```suggestion
    * (Procedural-style; use `$index` parameter): `civicrm_api4('Contact', 'get', [], 'id')`
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+    - (OOP-style; use `indexBy()` method): `\Civi\Api4\Contact::get()->execute()->indexBy('id')`
```
MikeyMJCO:
```suggestion
    * (OOP-style; use `indexBy()` method): `\Civi\Api4\Contact::get()->execute()->indexBy('id')`
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Input
+- Instead of a single `$params` array containing a mishmash of fields, options and parameters, each APIv4 parameter is distinct (see section on Params below).
```
MikeyMJCO:
```suggestion
* Instead of a single `$params` array containing a mishmash of fields, options and parameters, each APIv4 parameter is distinct (see section on Params below).
```

docs/api/v4/differences-with-v3.md
```diff
@@ -0,0 +1,166 @@
...
+- Custom fields are refered to by name rather than id. E.g. use `constituent_information.Most_Important_Issue` instead of `custom_4`.
```
MikeyMJCO:
```suggestion
* Custom fields are refered to by name rather than id. E.g. use `constituent_information.Most_Important_Issue` instead of `custom_4`.
```

mkdocs.yml
```diff
@@ -54,12 +54,11 @@ pages:
   - APIv4:
     - APIv4 Usage: api/v4/usage.md
     - APIv4 Actions: api/v4/actions.md
-  #  - APIv4 Options: api/v4/options.md
     - APIv4 Joins: api/v4/joins.md
     - APIv4 Chaining: api/v4/chaining.md
     - APIv4 Custom Data: api/v4/custom-data.md
-  #  - APIv4 Examples: api/v4/examples.md
-  #  - APIv4 Changes: api/v4/changes.md
+    - Differences between Api v3 and v4: api/v4/differences-with-v3.md
```
MikeyMJCO: Can we move this to the "root" level of the API folder and drop the v4 part of the name. e.g:

```suggestion
    - Differences Between API v3 and v4: api/differences-between-API-v3-v4.md
```

colemanw: IMO this section is fine where it is because it's really about v4 and how to upgrade from previous versions; it's not about the api in general.

MikeyMJCO: That's fair - still need to correct for "Title" case in the page title maybe? Otherwise I'm good with this!

colemanw:
Ok thanks @MikeyMJCO - feel free to improve that table if you want :)

PR #715, opened 2019-10-24
corrected extension name

amirasyraf:
The extension name is just `myextension`, not `com.example.myextension`.

See https://chat.civicrm.org/civicrm/channels/dev-newcomers


MikeyMJCO:
Thanks for this, the example is however correct - we cannot yet fully deprecate the long name syntax. Depending on the extension the longer command type could work or even be required.

MikeyMJCO:
In the context of Civix (developing new extensions) this change makes sense.

PR #716, opened 2019-10-24
Document BuildKit EoL Policy

MikeyMJCO:
Implements discussion which took place in civicrm/civicrm-buildkit/issues/432

### DIFF
--- a/docs/tools/buildkit.md
+++ b/docs/tools/buildkit.md
@@ -6,42 +6,6 @@ Many of these tools are commonly used by web developers, so you may have already
 
 This is the same collection of tools which manages the test/demo/release infrastructure for civicrm.org.
 
-## Supported Operating Systems
-
-Currently Buildkit includes [specific, tested install steps](https://github.com/civicrm/civicrm-buildkit/blob/e278123c2a5edb525a01c816283d30b882aa8fab/bin/civi-download-tools#L329) for the following Ubuntu and Debian operating system releases - note that recently removed versions are shown in this list for information and are marked in the final column.
-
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
@MikeyMJCO so I find this confusing as a mac & occassionally windows buildkit user. It implies there is no support for either. For windows I have only used in vagrant so that might be true but for Mac there are no obvious support issues

MikeyMJCO:
> @MikeyMJCO so I find this confusing as a mac & occassionally windows buildkit user. It implies there is no support for either. For windows I have only used in vagrant so that might be true but for Mac there are no obvious support issues

Pushed a change to address - better?

eileenmcnaughton:
Yes - that looks good @MikeyMJCO - maybe @demeritcowboy has more comments re buildkit on a mac

demeritcowboy:
str_replace('mac', 'windows', $this->getPreviousComment());

MikeyMJCO:
lol :-D

I'm gonna wait for review by @totten on this one before merging :-)

totten:
@MikeyMJCO I agree it's useful to give a specific list of supported/unsupported versions for people looking at boundary cases. (*Is my version from Ubuntu X.10 too new because it just came out? Is my EOL release still OK?*) But the full table feels like... a lot of information that most readers don't need to read. The TLDR is that release should be *recent/supported*.

I pushed a proposed update. What do you think of tweaking the text to put more emphasis on recent/supported and using a hyperlink to define?

MikeyMJCO:
Yeah - that's better - stick it as an appendix at the end of the page - good calL!

seamuslee001:
changes are ok by me

PR #717, opened 2019-10-24
Update explorer url

colemanw:
The inconsistency was bugging me so I created https://github.com/civicrm/civicrm-core/pull/15597.

But we may want to wait to merge this until that PR actually makes it into a release.


MikeyMJCO:
This will drop in 5.20 right?

seamuslee001:
AFAIK yes

seamuslee001:
Merging this as 5.20 has gone out now

PR #718, opened 2019-10-24
Improve APIv4 usage docs

pfigel:
This contains a mix of typo and style fixes, removes some bits that were copied from APIv3 and adds or amends some sections.


totten:
:+1: Agree, looks better.

PR #720, opened 2019-10-25
Nest "Create the default payment processor file" section under "Setup"

alifrumin:
It seems to me like  "Create the default payment processor file" should be the first subheading under "Setup"

### DIFF
--- a/docs/extensions/payment-processors/create.md
+++ b/docs/extensions/payment-processors/create.md
@@ -66,14 +66,14 @@ Checkout - the info is entered at the processors' site.
 
 
 
-### Create the default payment processor file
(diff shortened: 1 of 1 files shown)


MikeyMJCO:
Seems ok - but as you're touching it - can you change it so that the title is in Title Case :-p

Tag - you're it ;-)

alifrumin:
Hopefully that helps! I nested the "Add any additional libraries needed" section or rather the "Add Any Additional Libraries Needed" section too as that seems to fall into setup.

MikeyMJCO:
Thanks Alice :-)

PR #725, opened 2019-10-29
Document regen.sh

demeritcowboy:
Some documentation for bin/regen.sh


colemanw:
Definitely better than what was there before :)

PR #726, opened 2019-10-29
Formatting improvements and add code comments on Order API + fix broken URL

artfulrobot:
- Fix broken link to https://lab.civicrm.org/dev/financial/issues/73
- Use `//` for single line comments (isn't `#` deprecated? or is that just in .ini files?)
- Add whitespace and comments around migration from transact code.
- Wrap code example lines so they display properly

### DIFF
--- a/docs/financial/orderAPI.md
+++ b/docs/financial/orderAPI.md
@@ -35,8 +35,8 @@ Call `Order.create` with a structure like the below. Note that we always create 
   "contribution_status_id": "Pending",
   "line_items" : [
     {
+      "params": { },
       "line_item": [
-        "params": { },
         {
           "qty": 1,
           "unit_price": 1.23,
--- a/docs/financial/orderAPI.md
+++ b/docs/financial/orderAPI.md
@@ -359,8 +359,10 @@ try {
 
   // Assuming the payment was taken, record it which will mark the Contribution
   // as Completed and update related entities.
-  civicrm_api3('Payment', 'create',
-    ['contribution_id' => $order['id'], 'amount' => $params['amount']]);
+  civicrm_api3('Payment', 'create', [
+    'contribution_id' => $order['id'],
+    'total_amount' => $params['amount'],
+    ]);
 }
 catch  {
   // it failed
--- a/docs/financial/orderAPI.md
+++ b/docs/financial/orderAPI.md
@@ -62,7 +62,7 @@ Things to note:
 5. The `line_total` *must* equal the `unit_price` × `qty`
 
 !!! info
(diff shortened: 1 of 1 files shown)


docs/financial/orderAPI.md
```diff
@@ -351,8 +353,16 @@ if (!isset($params['invoiceID']) {
...
+  // Use the Payment Processor to attempt to take the actual payment. You may
+  // pass in other params here, too.
   civicrm_api3('PaymentProcessor', 'pay', ['contribution_id' => $order['id']]);
-  civicrm_api3('Payment', 'create', ['contribution_id' => $order['id'], 'amount' => $params['amount']]);
+
+  // Assuming the payment was taken, record it which will mark the Contribution
+  // as Completed and update related entities.
+  civicrm_api3('Payment', 'create', [
+    'contribution_id' => $order['id'],
+    'total_amount' => $params['amount'],
```
eileenmcnaughton: I've realised we could probably deprecate not passing paymen_processor_id or payment_instrument_id in here - perhaps we should add that to the example

artfulrobot: So you mean we *should* insist on passing in `payment_processor_id` and `payment_instrument_id` ?

eileenmcnaughton: OR not AND  - we can figure out the payment_instrument_id if we have the payment_processor_id

eileenmcnaughton: but yes - I think we can't really add payments without knowing what sort of payments they are without too much magic & payment method should be known when creating a payment

artfulrobot: I've added both in because that's the most efficient way (saves a lookup on instrument). It currently looks up intstrument from the contribution too, if processor not given, but anyway.

I wonder if `payment_processor_id` should be required for `Payment.create` - it seems really important and is stored in the financial transaction table.

eileenmcnaughton: Well processor_id is only relevant for processor payments -not checks. If you pass in payment_instrument_id & it is not the same as the one associated with they payment processor I think it should be ignored - meaning a  look up is still needed

eileenmcnaughton:
@artfulrobot ok I logged that here
https://lab.civicrm.org/dev/financial/issues/95

@MikeyMJCO I think this is OK to merge

docs/financial/orderAPI.md
```diff
@@ -335,11 +335,13 @@ The Contribution.transact api will create a 'simple' contribution and process a
...
 The simplest first step to migrate off it is to replace the order api call with a call that follows the recommended flow but still does not address the line item creation gaps & it is recommended you  look at the patterns above to do that. This first step looks like
 
-```
-# start with the same parameters as Contribution.transact.
+```php
+// Start with the same parameters as Contribution.transact.
 $params = $transactParams;
-# it would be  better just  to include the relevant params but....
+
+// It would be  better just  to include the relevant params but....
```
MikeyMJCO:
```suggestion
// It would be better just to include the relevant params but....
```

MikeyMJCO:
Approved - merging!

eileenmcnaughton:
Yay thanks - great this documentation is progressing

PR #732, opened 2019-11-16
tools/issue-tracking.md - Update guidance on Gitlab labels

totten:
__Before__: The documentation explains ~5 common labels.  Gitlab actually has ~50
labels.

__After__: The documentation explains ~3 important labels and ~4 *prefixes*.
Each prefix is given with a definition and example, and there are links
to the full lists.

__Comment__: The link goes to https://lab.civicrm.org/dev/core/-/labels
though arguably https://lab.civicrm.org/groups/dev/-/labels might be better.
However, there a dozen or so labels which (rightly or wrongly) only show up
in `dev/core`, so `dev/core` gives the more comprehensive listing.

--

@eileenmcnaughton @bjendres @jaapjansma - This is the corresponding docs update for that table that I drafted on MM. Note that there is one stylistic change - when something has a component name ("CiviMail", "CiviMember"), it uses that name (`comp:CiviMail` instead of `comp:mailings`). It's kind of a toss-up on aesthetic grounds, but I think this label is more forgiving when the user does data-entry:

Screen Shot 2019-11-15 at 5 36 07 PM

| If the user types... | and if the label is... | then the user gets the right label |
| -- | -- | -- |
| "mail" | `comp:mailings` | Yes, matches
| "CiviMail" | `comp:mailings` | No
| "mail" | `comp:CiviMail` | Yes, matches
| "CiviMail" | `comp:CiviMail` | Yes, matches

### DIFF
--- a/docs/tools/issue-tracking.md
+++ b/docs/tools/issue-tracking.md
@@ -183,25 +183,25 @@ Triage is done from this screen https://lab.civicrm.org/groups/dev/-/boards and 
 
 ### Categorisation
 
-During triage the following labels may be added:
+In Gitlab, issues are categorized by "Label", and multiple labels may be applied.  A few of the most important labels are standalone terms:
 
-_Critical_
-If the bug is a critical bug this label will be added. UPDATE- we don't really add this any more since by definition if a bug has been present for the entire 5.x series it's not critical. It would need to be something really big to get this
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
Seems OK to me

jaapjansma:
Seems also ok to me!  Thanks for this update and cleaning up of the labels.

bjendres:
Sounds good to me, too!

PR #735, opened 2019-11-25
Improve create payment processor documentation

artfulrobot:
## Before

Really hard to write a payment processor; results in wild variations between processors; results in difficult to maintain code and inconsistencies.

## After

Include documentation on the responsibilities and roles of the various payment processor classes; what data and logic they should handle and what they should not.

## PR WIP

What do you think about this? @eileenmcnaughton @mattwire 

I'm desperately trying to do things right while keeping things working on the GoCardless extension, but every example I find to reference seems riddled with inconsistencies, which I suspect is to do with things being written at different points in time; guesswork; bodges/hacks for particular awkward third parties; changes in core (like Order API)... The existing documentation has loads of problems IMO, so rather than just sigh I thought I'd have a stab at setting stuff straight.

But this may not be the best way to do it; I don't know, so don't hold back if you think there's a better way to improve things.


mattwire:
@artfulrobot You should find quite a bit of useful commenting in the latest stripe / mjwshared extensions. In general I've tried to "do things the right way" in the stripe extension and handle all the required hacks / bodges in the mjwshared extension. I've tried to comment with reasons why in the code

docs/extensions/payment-processors/create.md
```diff
@@ -1,5 +1,227 @@
...
+
+Because things get complex and because `CRM_Core_Payment` is a bridge between CiviCRM's logic and the payment processor service's needs, it's all too easy to end up combining business logic (like updating membership end dates, or deciding whether to send a receipt email) with your calls to the external service. If you do need to do this try to separate out your class methods.
+
+CiviCRM's Contribution and Event pages are obvious users of your `CRM_Core_Paymnet` class but you should not assume that those are the only consumers; it should be able to be used by other processes too, e.g. Drupal webform or an entirely bespoke process. **Therefore they should not call functions that assume a user context** such as redirects, exits, or setting status messages like `CRM_Core_Error::statusBounce`.
```
JoeMurray: I'm highly concerned about any approach that suggests regularly overriding a core file of any kind, but especially something as complex and touchy as CRM_Core_Payment. I'd prefer to recommend that before any payment process tries overriding that file that they discuss their needs with the community to try to figure out a way to make core serve their needs.

docs/extensions/payment-processors/create.md
```diff
@@ -1,5 +1,227 @@
...
+```
+
+This means the value is still validated and type-cast as an amount (in this example).
+
+### Custom payment processor-specific data
+
+!!! danger "todo"
+    This is not yet implemented in core.
+
+Sometimes a payment processor will requrie custom data. e.g. Stripe may use a `paymentMethodID` and a `paymentIntentID` on its `doPayment()` method.
```
JoeMurray: Hmm. I'm a bit confused by this as core has a payment method id.

docs/extensions/payment-processors/create.md
```diff
@@ -1,5 +1,227 @@
...
+The payment class is responsible for validating such data; anything is allowed by `setCustomProperty`, including `NULL`.
+
+Where support for custom properties is needed for a method, e.g. `doPayment()`, you should implement a method as follows:
+
+```php
+<?php
+use Civi\Payment\PropertyBag;
+use Civi\Payment\Exception\PaymentProcessorException;
+
+class CRM_Core_Payment_MyExtension extends CRM_Core_Payment {
```
JoeMurray: I don't mind extending CRM_Core_Payment, it's the overriding of its methods or properties that I would be concerned about, insofar as they don't invoke parent's or change the results of invoking parent's.

docs/extensions/payment-processors/create.md
```diff
@@ -1,5 +1,227 @@
...
+  /**
+   * Copy our custom properties from the form data.
+   *
+   * @param PropertyBag
+   * @param Array form values
+   * @param String $component as for doPayment()
+   */
+  public function extractCustomPropertiesForDoPayment(PropertyBag $propertyBag, Array $input, $component = 'contribute') {
+
+    // We require a paymentIntentID
```
JoeMurray: Just for my info (and maybe others'), it might be useful to explain above what a payment Intent is.

docs/extensions/payment-processors/create.md
```diff
@@ -1,5 +1,227 @@
...
+- `contributionID` (See [dev/financial#53](https://lab.civicrm.org/dev/financial/issues/53))
+- `amount` e.g. *1.23*
+- `isRecur`
+- `email` A billing address *foo@example.com* that should be used by the external processor (if it uses it at all)
+
+!!! danger
+    The following is a proposal
+
+
+I propose that the following changes are made throughout core: old: `$paymentProcessor->doPayment($params);` new:
```
JoeMurray: @eileenmcnaughton once you're back in February I think this is a place where you should weigh in.

docs/extensions/payment-processors/create.md
artfulrobot: @JoeMurray yes, I agree, and the fact that you thought I didn't from the text above means it needs revision!

However, to be clear, the job of someone writing a payment processor class is exactly to override `CRM_Core_Payment` through inheritance - we're not replacing `CRM_Core_Payment` but that class is not called itself; our subclass is.

Some processors I've seen do too much - they do it out of desperation not desire - like handling membership stuff with their own logic instead of leaving it to the correct APIs. Shifting APIs and lack of documentation are a barrier to that, which is part of my motivation for making these documentation updates/additions.

Can you suggest a rewording?

docs/extensions/payment-processors/create.md
artfulrobot: Well, core has various names and aliases of names (Financial Instrument, Payment method etc) I was giving a real world example where Stripe (the 3rd party company's API, not the extension by Matt) requires a special data field or two - the names are Stripe's names and any similarity to CiviCRM's names is coincidental.

My proposal is that all custom data requirements have their properties prefixed, therefore these would become `stripe_paymentMethodID` and `stripe_paymentIntentID` to help avoid the ambiguity.

docs/extensions/payment-processors/create.md
artfulrobot: This is something that I've suggested we have a voice discussion about.

docs/extensions/payment-processors/create.md
artfulrobot: Sure. I'll do that. I thought that using a real world example would be more helpful than using `someCustomProperty` or `magicSpecialCode` or something like that but maybe it's not clear. I'll have another go at editing. Thanks @JoeMurray

docs/extensions/payment-processors/create.md
artfulrobot: Quite. But's not a case of "don't mind" - creating a new payment processor *requires* extending that class. But 2 things are:

1. we should work within it, overriding the only bits we're supposed to, and ensuring that we respect the interfaces in terms of input and output types.

2. we should be able to achieve our aims without needing to do (1)!

docs/extensions/payment-processors/create.md
artfulrobot: I've done a new commit with lots of rewording - hopefully it's clearer now.

docs/extensions/payment-processors/create.md
artfulrobot: I've done a new commit with some rewording, hopefully clearer now, but comment on that if not. Thanks.

docs/extensions/payment-processors/create.md
artfulrobot: I've done rewording here, pls. comment on that if still unclear. Thanks.

docs/extensions/payment-processors/create.md
JoeMurray: Sorry, it's much clearer now that I read it over again, saying it extends rather than overrides. Maybe mention implementing stub classes or abstract interfaces? There should not be interactions with business logic eg for handling memberships..that's a clear case of a problem. If necessary core needs changes but more likely new developer needs to be shown right way or workaround way.

artfulrobot: I've tried to work in your concerns, if you want to take a look at the latest.

docs/extensions/payment-processors/create.md
JoeMurray: My bad.

docs/extensions/payment-processors/create.md
MikeyMJCO: I don't think proposals belong in docs content.

MikeyMJCO:
@artfulrobot what's the status/likely movement on this?

MikeyMJCO:
@artfulrobot thanks for this - I'm going to merge this as-is and we can improve incrementally from here :-)

PR #739, opened 2019-12-08
Add in documentation on setting and retrieving contact settings using…

seamuslee001:
… the Civi:: facade and also add in note about the deprecation of the ::setItem function

ping @eileenmcnaughton @totten

### DIFF
--- a/docs/framework/setting.md
+++ b/docs/framework/setting.md
@@ -25,7 +25,9 @@ The "Settings" system developed incrementally:
 - v4.2 – Added public Settings API. Included support for CRUD'ing data from multiple sources (`civicrm_domain` and `civicrm_setting`). Migrated more settings data from `civicrm_domain` to `civicrm_setting`.
 - v4.3 to v4.6 – Incrementally migrate more settings data.
 - v4.7 – Added `Civi::settings()` and refactored init system. Finished migrating data from `civicrm_domain` to `civicrm_setting`.
+- v5.7 - Added `Civi::contactSettings()` as a facade to allow for managing of Contact Settings rather than domain level settings.
 - v5.8 - Added Generic CiviCRM settings form.
+- v5.21 - Deprecated the usage of `CRM_Contact_BAO_Setting::setItem` in favour of the `Civi::settings` and `Civi::contactSettings` functionality and the Setting API to set settings
 
 ## Settings Definition
 
@@ -90,7 +92,13 @@ Deprecated settings properties are :
 
 ## Setting Storage and Retrieval
 
-To set a setting, developers can use either the Setting.Create API or use `Civi::settings()->set()` to set the value e.g. `Civi::settings()->set('editor_id', 'CKEditor');`. To Retrieve a setting value, developers can use the Setting.Get API or use `Civi::settings()->get()` e.g. `Civi::settings()->get('max_attachments');`
+### Domain Settings
+
+To set a setting for a whole domain, developers can use either the Setting.Create API or use `Civi::settings()->set()` to set the value e.g. `Civi::settings()->set('editor_id', 'CKEditor');`. To Retrieve a setting value, developers can use the Setting.Get API or use `Civi::settings()->get()` e.g. `Civi::settings()->get('max_attachments');`
+
+### Contact Settings
+
+After 5.7.0 developers are able to use the `Civi::contactSettings` facade to set contact specific settings. You can also use the API to set it but must pass in a contact_id parameter. To set a contact setting it would be like `Civi::contactSettings($contact_id)->set('navigation', $value)`. To retrieve the setting developers can use `Civi::contactSettings($contact_id)->get('navigation')`
 
 ## API actions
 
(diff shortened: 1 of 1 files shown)


docs/framework/setting.md
```diff
@@ -25,7 +25,9 @@ The "Settings" system developed incrementally:
 - v4.2 – Added public Settings API. Included support for CRUD'ing data from multiple sources (`civicrm_domain` and `civicrm_setting`). Migrated more settings data from `civicrm_domain` to `civicrm_setting`.
 - v4.3 to v4.6 – Incrementally migrate more settings data.
 - v4.7 – Added `Civi::settings()` and refactored init system. Finished migrating data from `civicrm_domain` to `civicrm_setting`.
+- v5.7 - Added `Civi::contactSettings()` as a facade to allow for managing of Contact Settings rather than domain level settings.
 - v5.8 - Added Generic CiviCRM settings form.
+- v5.21 - Deprecated the usage of `CRM_Contact_BAO_Setting::setItem` in favour of the `Civi::settings` and `Civi::contactSettings` functionality and the Setting API to set settings
```
eileenmcnaughton: No - this was discouraged long long before 5.21 - we've told people to use the api since about 4.2

eileenmcnaughton: (saying deprecated here implies it was ever encouraged or supported for external use)

eileenmcnaughton:
@seamuslee001 @totten so before we merge this I'd like to talk about the serialisation handling.

apiv3 unserialises the settings to an array. It either does, or should, recognises metadata with the serialisation type. I believe the Civi::setting facade doesn't - so I find it can be used for strings but not in various other places.

I feel like it should - but if it's too hard to add we should handle it as an api v4 setting api & make that the one encouraged method

seamuslee001:
@eileenmcnaughton are you saying that you would want some settings to be say able to store a json blob or something?

docs/framework/setting.md
seamuslee001: would something like issued e-notices or something instead of deprecated make more sense?

eileenmcnaughton:
@seamuslee001 right a key like  'user_dashboard_options' comes in & out of the apiv3 as an array - you don't need to know that it is serialised or how. But in fact the spec is

  ```  'serialize' => CRM_Core_DAO::SERIALIZE_SEPARATOR_BOOKEND,```

and if you pass in an array of options the pseudoconstant will help you out too.

The spec expects that the serialize key works much the same as for the DAO (although I have a feeling there might be a bug there) but I believe that we want the recommended interface to accept an array for fields like the above. And if that is not the Civi::setting facade then it probably should be a v4 api.

Having said that it is probably safe to retrofit to Civi::setting as the difference between receiving an array & a string can be tested

docs/framework/setting.md
eileenmcnaughton: I suppose we could say something like 
CRM_Contact_BAO_Setting::setItem has not been supported from outside the core code base for some years but as of 5.21 it will issue notices in preparation for removal in a subsequent version

eileenmcnaughton: (actually it has never been supported I would argue)

totten:
> apiv3 unserialises the settings to an array. It either does, or should, recognises metadata with the serialisation type. I believe the Civi::setting facade doesn't - so I find it can be used for strings but not in various other places.

I think the latter bit is incorrect. Try it out:

```
$ cv ev 'Civi::settings()->set("foo", [2, 4, 8]);'

$ cv ev 'var_export(Civi::settings()->get("foo"));'
array (
  0 => 2,
  1 => 4,
  2 => 8,
)

$ echo 'select value from civicrm_setting where name = "foo";'  | cv sql
value
a:3:{i:0;i:2;i:1;i:4;i:2;i:8;}
```

The settings table is a bit different from other tables - as far as I can recall, it has always used the PHP serialize format to store *all* settings, and it's always been best to access it through some helper (whether that's `civicrm_api()` or `CRM_Core_BAO_Setting::getItem()` or `Civi::setting()`) which hides this. 

The broader situation is more ambiguous -- e.g. some tables might store data in one format, and others in a different format. So that's a challenge for APIv3/APIv4 since they have generic CRUD that works on all entities. But `Civi::settings()` only deals with the one table, and it should be correct in that.

docs/framework/setting.md
totten: OK, so it was nominally deprecated, and 5.21 will start to emit warnings. So suggestion:

```suggestion
- v5.21 - `CRM_Contact_BAO_Setting::setItem()` will now emit deprecation warnings. Use `Civi::settings()`, `Civi::contactSettings()`, or the Setting API instead.
```

eileenmcnaughton:
hmm maybe the issue I'm thinking of is the difficulty of wrangling options within the array - I'll have to double check. I know I abandoned it recently on something like that.

seamuslee001:
ok @eileenmcnaughton @totten i think this is right now @MikeyMJCO

PR #741, opened 2020-01-06
Add better example for transaction handling

mattwire:
@totten


totten:
@mattwire Yup, agree that's better advice. :+1:

PR #746, opened 2020-01-12
Updates requirements

stesi561:
For php and mysql based on https://docs.civicrm.org/sysadmin/en/latest/requirements/#hosting

### DIFF
--- a/docs/basics/requirements.md
+++ b/docs/basics/requirements.md
@@ -3,8 +3,8 @@
 ## Languages and Services
 
 -   Unix-like environment (Linux, OS X, or a virtual machine)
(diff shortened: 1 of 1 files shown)


stesi561:
This possibly needs to be incremented to 7.1 next month? 
Not sure about other versions of things.

MikeyMJCO:
@totten @seamuslee001 @eileenmcnaughton

This was being discussed on GitLab - how do we feel about changing the docs in this way?

eileenmcnaughton:
@MikeyMJCO so this seems like it only gets us to being accurate about what is currently  supported  - so good to merge. Step 2 is to update again once 7.0 is gone...

PR #750, opened 2020-01-20
Reorganze Api4 Usage

colemanw:
This page was kind of inconsistent. Some of the information was way too basic (like what is a Contact - that's clearly covered elsewhere), and some was incredibly esoteric (like how to hack the Api by invoking the kernel directly - tip: don't)

I've tried to reorganize it to be approachable and useful.


MikeyMJCO:
Change looks good to me - separate note but can you make changes in a fork rather than the main repo - creating branches in this repo creates a ton of log noise on docs-publisher at the moment :-/ until I can fix it properly we should try and avoid branches in the docs repos.

colemanw:
Ok @MikeyMJCO will do that next time.

PR #754, opened 2020-01-30
Note about when to consider cost vs. benefit of "ts()"

demeritcowboy:
Separated out from https://github.com/civicrm/civicrm-dev-docs/pull/751

Sometimes if the wording itself is fine and the technical functionality is working and it's not confusing users, having the exact right code isn't worth invalidating an existing translation and generating another end-to-end cycle of the translation process.

### DIFF
--- a/docs/translation/index.md
+++ b/docs/translation/index.md
@@ -205,6 +205,25 @@ Typography is different in different languages and thus must be translated along
     {ts}Event Total:{/ts}
     ```
 
+### Consider leaving existing bad uses of ts unchanged when the intended functionality is working
+
+Because of demands on translators' time and the number of steps in the end-to-end translation process itself, it is sometimes better to leave existing bad instances of `ts` unchanged.
+
+An example that you should consider leaving as is would be:
+
+!!! failure "Consider leaving this as-is"
+
+    ```smarty
+    {ts 1="https://www.example.org"}This is a block of text where the wording is fine and the link works but the <a href="%1" target="_blank">link attributes</a> should be a parameter to the ts. But it's not preventing anyone completing their work tasks and isn't confusing to users.{/ts}
+    ```
+
+The above `ts` should be written as `{ts 1='href="https://www.example.org" target="_blank"'}` and then the link should be `<a %1>`, but making that change will break the existing translation just for some non-essential technical rearrangement. Within two CiviCRM release cycles the string will suddenly have become untranslated and it may be several months after that before it appears retranslated in a release.
+
+!!! success "Examples that should be changed"
+
+* If it would normally break translation but the new string is already used somewhere else, and so it's already translated.
+* Spelling mistakes. While not preventing a user from completing a task, the positives outweigh the negatives, and this is visible to the user, as compared to the above "don't change" example where the user doesn't see a problem.
+
 ## Rationale for using Gettext
 
 In most projects, strings are typically translated by either:


docs/translation/index.md
```diff
@@ -205,6 +205,16 @@ Typography is different in different languages and thus must be translated along
...
 
+### Consider leaving existing bad uses of ts unchanged when the intended functionality is working
+
+Because of demands on translators' time and the number of steps in the end-to-end translation process itself, it is often better to leave existing bad instances of `ts` unchanged unless they are preventing users from completing tasks. For example a link that is malformed html and so doesn't go anywhere when clicked should be changed. An example that you should consider leaving as is would be:
+
+!!! failure "Consider leaving this as-is"
+
+    ```smarty
+    {ts 1="https://www.example.org"}This is a block of text where the wording is fine and the link works but the <a href="%1" target="_blank">link attributes</a> are wrong. But it's not preventing anyone completing their work tasks and isn't confusing to users.{/ts}
+    ```
```
colemanw: It took me a while to understand this example. I think by "link attributes are wrong" you mean like `target="_blank"` needs to be changed.

If this were a real-world example, as a developer this one would be a tough call for me. Leaving it alone is an option, but if it needed to be changed I'd try to make a swift change that avoids future pain, so I'd probably change the code to:

    ```smarty
    {ts 1='href="https://www.example.org" target="_something_else"'}This is a block of text where the wording is fine and the link works but the <a %1>link attributes</a> are wrong. But it's not preventing anyone completing their work tasks and isn't confusing to users.{/ts}
    ```

That way attributes can be updated in the future without changing the gettext string. While I was at it I'd carefully review the wording, since the string is impacted by this change anyway, and make any nitpicky tweaks that we normally wouldn't bother translators with.

demeritcowboy: I mean everything is working from the user point of view and the wording is fine, but the target should be in the ts parameter and not have one attribute in the ts and one in the `a`. But when I did that kind of cleanup in real life see paragraph 3 at https://github.com/civicrm/civicrm-core/pull/16381#issuecomment-578422899. So let's make a policy either don't clean these up, or take the hit and do clean it up.

demeritcowboy: I see what you mean by "wrong" not being clear. I've made an update to try to clarify the clarification.

MikeyMJCO:
@totten @eileenmcnaughton @seamuslee001 @colemanw (and probably others!)

I think we're having a policy debate in a PR here - can we clarify what our policy is going forward for these kinds of changes and we'll document based on that (or just put that in the documentation if short enough!)

demeritcowboy:
I'm ok to drop and move this out of the queue if it's too hard to make guidelines for or if there is no agreement on the guidelines. How about I'll leave it for a few more days in case anyone wants to comment and then I'll close it if it's not going anywhere.

MikeyMJCO:
> I'm ok to drop and move this out of the queue if it's too hard to make guidelines for or if there is no agreement on the guidelines. How about I'll leave it for a few more days in case anyone wants to comment and then I'll close it if it's not going anywhere.

Sounds good - but I want the discussion in any event - if needed we'll move it to a GitLab issue but I'm okay with it here for a bit!

wmortada:
This sounds like a useful point to raise. I confess my ignorance about how the translation system works so it would be useful to have some clear guidance about what should and shouldn't be changed.

My personal preference would be to correct all spelling mistakes because they really stand out and bother me and I think it makes the software look unprofessional. However, I appreciate that this may come with an impact on translations and would be keen to understand how to minimise this.

MikeyMJCO:
I'm going to move this to a GitLab issue in the next few days and close out the PR.

demeritcowboy:
Yeah let's move it to gitlab. I'll make a ticket.

demeritcowboy:
Moved to https://lab.civicrm.org/dev/translation/-/issues/42

PR #755, opened 2020-01-30
Note about title attributes in anchor tags

demeritcowboy:
Separated out from https://github.com/civicrm/civicrm-dev-docs/pull/751

This might not make complete sense on its own without 751, but it seems difficult to get the wording for this part right so separating it out.


docs/translation/index.md
```diff
@@ -99,7 +99,29 @@ The general rules for avoiding errors may be summed up like this:
...
+
+    ```smarty
+    {ts}<a href="https://www.example.org/civicrm/contact/view?cid=1" title="Go to the contact page">View Contact</a>{/ts}
+    ```
+
+!!! failure "Less bad"
+
+    ```smarty
+    <a href="https://www.example.org/civicrm/contact/view?cid=1" title="{ts}Go to the contact page{/ts}">{ts}View Contact{/ts}</a>
+    ```
```
colemanw: I'm not sure I get why this is still sorta bad. Because the title & link text are redundant?

MikeyMJCO: Yes - I think it needs a better example perhaps?

demeritcowboy: I was basing the example on existing defacto usage and I'm not clear myself whether there is a "good" way to do title attributes. I put "less bad" because you don't have the context that they go together.

demeritcowboy: Updated example to one where the title text is more different from the link text.

PR #756, opened 2020-01-30
Fix table display

colemanw:
I guess this will work for now...


colemanw:
@MikeyMJCO I guess this is the best we can do pending theme updates? It should at least fix the broken-ness of that page.

MikeyMJCO:
I agree! Definitely go ahead for now and we can revisit this page when I make some theme changes - I may be able to get something done sooner-rather-than-later!

PR #757, opened 2020-01-30
Delete v3 api-architecture page

colemanw:


colemanw:
On second thought @MikeyMJCO made a good point that we could just delete this page.
It's out-of-date even for v3 and we have a shiny new one for v4 that is actually useful.

MikeyMJCO:
> On second thought @MikeyMJCO made a good point that we could just delete this page.
> It's out-of-date even for v3 and we have a shiny new one for v4 that is actually useful.

Unexpectedly I +100 this - realistically we shouldn't be publicising how to build on/with APIv3 any longer and this page is already out of date even if we ignore that fact!

colemanw:
PR updated @MikeyMJCO

PR #758, opened 2020-02-03
Remove pass-by-reference from object variable.

mfb:
Pass-by-reference notation isn't necessary for objects - this is probably a remnant from php 4? :)


colemanw:
While we're at it, should clean up the variable name because "object" is just sooo descriptive.
How about `$reportForm`

PR #760, opened 2020-02-14
Tweak api architecture & changelog and move hook changelog to its own page

colemanw:
The big change here is to move the hook changelog to its own page since it does not belong with APIv3.


colemanw:
@MikeyMJCO see if you agree with this reorganization

colemanw:
@eileenmcnaughton you might also have an opinion about moving hook changes to a separate page.

MikeyMJCO:
This looks very good to me in principle. I haven't done an in-depth content review but I'm okay with the reorganisation aspect!

eileenmcnaughton:
Makes sense I guess - I don't have a strong opinion here as long as it's easy to find

PR #761, opened 2020-02-15
Add documentation for tags

eileenmcnaughton:
I've kept this limited - if people agree we can merge this & then future PRs can add more as I think we probably want more recommended ones

Per https://github.com/civicrm/civicrm-core/pull/16551

### DIFF
--- a/docs/extensions/info-xml.md
+++ b/docs/extensions/info-xml.md
@@ -291,33 +291,47 @@ Elements acceptable within `<requires>`
 ### `<tags>` {:#tags}
 
 * Containing element: [`<extension>`](#extension)
-* Description: Freeform tags. By convention, tag-names should be "snake-case" (all lower-case alphanumeric with dashes for separate words). If namespaces or prefixes are required, tag-names may use a colon (`:`) delimiter. Tag-names should be documented on this page. (*If you need to use a new/unknown tag, then please submit an [update PR](https://docs.civicrm.org/dev/en/latest/documentation/) to this documentation with a description/purpose.*)
+* Description: Freeform tags allow extensions to be organized into categories or selected as subgroups. Tags MUST contain only letters, numbers, dashes,and colons. Tags are case-sensitive. Tags MUST follow the structure `<prefix>:<some-tag-name>`. Tags SHOULD be documented in the table below. (*If you need to use a new/unknown tag, then please submit a [documentation update](https://docs.civicrm.org/dev/en/latest/documentation/) with a description.*)
 * Contains: elements
 * Example:
 
     ```xml
     <extension key="org.civicrm.foo" type="module">
       <tags>
-        <tag>civicontribute</tag>
-        <tag>payment-processor</tag>
-        <tag>hidden</tag>
+        <tag>comp:CiviContribute</tag>
+        <tag>topic:payment-processor</tag>
+        <tag>mgmt:hidden</tag>
       </tags>
     </extension>
     ```
 
-The following tags are currently defined:
+The `topic:*` tags relate to a general topic/business area/problem area. Topics must use "snake-case" (lowercase/hyphenated).
 
-| Tag | Description |
-| --- | ----------- |
-| `hidden` | Do not display the extension in the administrative UI. (*Note: Pending implementation circa v5.24.*) |
-| `civicase` | The extension improves or alters CiviCase |
-| `civicontribute` |  The extension improves or alters CiviContribute |
-| `civigrant` | The extension improves or alters CiviGrant |
-| `civievent` | The extension improves or alters CiviEvent |
-| `civimail` | The extension improves or alters CiviMail |
-| `civireport` | The extension improves or alters CiviReport |
-| `civirules` | The extension improves or alters CiviRules |
-| `payment-processor` | The extension defines a payment-processor |
+| Tag name | Description |
+|--|--|
+| `topic:reporting` | Provides reporting functionality |
+| `topic:search` | Provides search functionality |
+| `topic:email` | Provides email functionality |
+
+The `comp:*` tags relate to specific implementations (components/extensions/modules). Components may use capitalization that matches the original component.  Where applicable, tags SHOULD match counterparts in the [Gitlab labels](https://lab.civicrm.org/dev/core/-/labels).
+
+| Tag name | Description |
+|--|--|
+| `comp:CiviCampaign` | Improves or alters the CiviCase component |
+| `comp:CiviCase` | Improves or alters the CiviCase component |
+| `comp:CiviContribute` | Improves or alters the CiviContribute component |
+| `comp:CiviEvent` | Improves or alters the CiviEvent component |
+| `comp:CiviGrant` | Improves or alters the CiviGrant component |
+| `comp:CiviMail` | Improves or alters the CiviMail component |
+| `comp:CiviReport` | Improves or alters the CiviReport component |
+
+The `mgmt:*` tags relate to the management of extensions. Management tags must use "snake-case" (lowercase/hyphenated).
+
+| Tag name | Description |
+|--|--|
+| `mgmt:hidden` | The extension is not displayed in the administrative UI, but it can be managed via CLI and API. (Implemented in v5.24+.) |
+| `mgmt:mandatory` | The extension must always be enabled. (*Implementation pending*) |
+| `mgmt:autoinstall` | The extension is installed by default on  new sites. (*Implementation pending*) |
 
 ### `<url>` {:#url}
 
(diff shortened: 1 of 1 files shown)


docs/extensions/info-xml.md
```diff
@@ -287,6 +288,33 @@ Elements acceptable within `<requires>`
 | -- | -- |
 | [`<ext>`](#ext) | 1+ |
 
+### `<tags>` {:#tags}
+
+* Containing element: [`<extension>`](#extension)
+* Description: Freeform tags -but not that there are guideline as to what to assign
```
mattwire: .. Freeform tags - but note that..

docs/extensions/info-xml.md
```diff
@@ -287,6 +288,33 @@ Elements acceptable within `<requires>`
...
+      <tags>
+        <tag>civicontribute</tag>
+        <tag>payment-processor</tag>
+        <tag>hidden</tag>
+      </tags>
+    </extension>
+    ```
+    
+* Notes:
+Tags may imply functionality and as that is developed that functionality / expectations will be documented here. It is not recommended to add tags without ensuring they are reflected here as functionality may be adding tags. These docs will be consulted at that point - but not any ad hoc tags in the wil.
```
mattwire: This needs rewording as it doesn't make sense to me right now.

totten:
I find myself vacillating about between different taxonomies.

* _Current draft_: `civicontribute`, `civicase`, `civimail`, `hidden`, etc
    * *I like that...* the words are fairly short, which is good for URLs and improvising CLI operations.
    * *I don't like that...* a majority of tags look like "civicontribute", which goes against the normal "CiviContribute" spelling.
* _Imitate the Gitlab/Github labels_: `comp:CiviContribute`, `comp:CiviCase`, `comp:CiviMail`, `ext:hidden`, etc
    * *I like that...* there's continuity across different properties (Gitlab/Github/extensions). Also, that this anticipates more dimensions (`comp:`, `sig:`, etc).
    * *I don't like that...* colons show up in pretty much everything.  
* _Aim for simple words; use entities/topics/themes instead of components_: `contributions`, `cases`, `mail`, `hidden`, etc
    * *I like that...* these tags are a bit broader - you might tag something as `mail` or `reports` even if it doesn't specifically use CiviMail or CiviReport. For example, `dataprocessor` could be tagged `reports` even though it's not `CiviReport`.
    * *I don't like that...* the relative flatness invites more improvising and possibility of future namespace conflicts.

eileenmcnaughton:
So, with some components it's implicit if the component is disabled so is the functionality - civicontribute, civievent - even frameworks like civireport (not yet disableable but one day), civirules

In that context saying that something is civireport tells us

1) that it extends CiviReport
2) that it adds reporting (ie it would show up on the 'reporting' section of whatever the extensions UI is)

Those 2 pieces of information are linked but would be used separately - one for the UI grouping & the other for 'should this be available'

docs/extensions/info-xml.md
```diff
@@ -291,7 +291,7 @@ Elements acceptable within `<requires>`
 ### `<tags>` {:#tags}
 
 * Containing element: [`<extension>`](#extension)
-* Description: Freeform tags -but not that there are guideline as to what to assign
+* Description: Freeform tags. By convention, tag-names should be "snake-case" (all lower-case alphanumeric with dashes for separate words). If namespaces or prefixes are required, tag-names may use a colon (`:`) delimiter. Tag-names should be documented in this file. (*If you need to use a new/unknown tag, then please submit an update PR to this documentation with a description/purpose.*)
```
twomice: Suggest changing "Tag-names should be documented in this _file_" to "Tag-names should be documented in this _page_". On first reading, I thought "file" was referring to the info.xml _file_, which is after all the topic of this _page_.

totten: Ah +1. That's better wording.

docs/extensions/info-xml.md
```diff
@@ -42,6 +42,7 @@ Here is an example of a full `info.xml` file from [CiviVolunteer](https://github
 
 | CiviCRM Version | Description |
 | -- | -- |
+|5.24.0|[`<tags>`](#tags) Tags introduced as part of schema|
```
MikeyMJCO:
```suggestion
| 5.24.0 | [`<tags>`](#tags) Tags introduced as part of schema |
```

docs/extensions/info-xml.md
```diff
@@ -287,6 +288,37 @@ Elements acceptable within `<requires>`
 | -- | -- |
 | [`<ext>`](#ext) | 1+ |
 
+### `<tags>` {:#tags}
+
+* Containing element: [`<extension>`](#extension)
+* Description: Freeform tags. By convention, tag-names should be "snake-case" (all lower-case alphanumeric with dashes for separate words). If namespaces or prefixes are required, tag-names may use a colon (`:`) delimiter. Tag-names should be documented in this page. (*If you need to use a new/unknown tag, then please submit an update PR to this documentation with a description/purpose.*)
```
MikeyMJCO:
```suggestion
* Description: Freeform tags. By convention, tag-names should be "snake-case" (all lower-case alphanumeric with dashes for separate words). If namespaces or prefixes are required, tag-names may use a colon (`:`) delimiter. Tag-names should be documented on this page. (*If you need to use a new/unknown tag, then please submit an [update PR](https://docs.civicrm.org/dev/en/latest/documentation/) to this documentation with a description/purpose.*)
```

totten:
> (@eileenmcnaughton) In that context saying that something is civireport tells us
>
> 1. that it extends CiviReport
> 2. that it adds reporting (ie it would show up on the 'reporting' section of whatever the extensions UI is)
>
> Those 2 pieces of information are linked but would be used separately - one for the UI grouping & the other for 'should this be available'

OK, if I'm understanding correctly, then these extensions would declare tags as:

| Extension key | Tags listed in `info.xml` | Implied tags |
| -- | -- | -- |
| `extendedreport` | `civireport` | `reporting` |
| `dataprocessor` | `reporting` | |

To which I'd offer two thoughts:

* I don't think the mechanism of *implying* buys a whole lot. If the list of tags are listed in a table in the dev-docs, then it's fairly easy for the developer to skim and pick the applicable ones. I'd keep it simpler with just explicit tagging.
* I think the namespace will get crowded if one is mixing tags which are structural (identifying the related technical component) as well thematic (identifying the business-area/topic).

If tagging can be used to flag both aspects, then probably it makes sense to use prefixes - like in the Gitlab/Github labels? So this might be a (partial) tag list:

| Tag name | Description |
|--|--|
| `comp:CiviReport` | Improves or alters the CiviReport component |
| `comp:CiviMail` | Improves or alters the CiviMail component |
| `topic:reports` | Provides reporting functionality |
| `topic:mail` | Provides email functionality |
| `mgmt:hidden` | The extension is not displayed in the administrative UI, but it can be managed via CLI and API |
| `mgmt:mandatory` | The extension must always be enabled. |
| `mgmt:autoinstall` | The extension is installed by default on  new sites. |

With the prefixes being:

| Prefix | Description |
| -- | -- |
| `comp:*` | Relates to a specific component/extension/implementation |
| `topic:*` | Relates to a general business/topic/problem area |
| `mgmt:*` | Relates to the administration of extensions |

Returning to do some examples with this taxonomy, we might ave:

| Extension key | Tags listed in `info.xml` |
| -- | -- |
| `extendedreport` | `topic:reporting`, `comp:CiviReport` |
| `dataprocessor` | `topic:reporting` |
| `sequentialcreditnotes` | `comp:CiviContribute`, `mgmt:hidden`, `mgmt:autoinstall` |

eileenmcnaughton:
I'm happy with the proposal ^^

totten:
@eileenmcnaughton @MikeyMJCO @twomice @mattwire - I've updated the draft to match.

docs/extensions/info-xml.md
```diff
@@ -287,6 +288,51 @@ Elements acceptable within `<requires>`
...
 
+### `<tags>` {:#tags}
+
+* Containing element: [`<extension>`](#extension)
+* Description: Freeform tags allow extensions to be organized into categories or selected as subgroups. Tags MUST contain only letters, numbers, dashes,and colons. Tags are case-sensitive. Tags MUST follow the structure `<prefix>:<some-tag-name>`. Tags SHOULD be documented in the table below. (*If you need to use a new/unknown tag, then please submit a [documentation update](https://docs.civicrm.org/dev/en/latest/documentation/) with a description.*)
+* Contains: elements
+* Example:
+
+    ```xml
+    <extension key="org.civicrm.foo" type="module">
```
eileenmcnaughton: we don't recommend this style of key now do we?

totten: All the other examples in the page use the Java style. I literally copied that line from the example right above. ✂️📋 

AFAIK, we've entered a grey transitional period where several people like the aesthetics of simpler names, and they seem to work in common cases, but no one can quite certify that  simpler names *always* work.

eileenmcnaughton: ok - let's ignore for now

eileenmcnaughton:
@MikeyMJCO  - I'm with merge & follow up if anything more

PR #765, opened 2020-02-23
dev/core#1615 - Import civicrm-setup docs

totten:
Per [dev/core#1615](https://lab.civicrm.org/dev/core/issues/1615), the `civicrm-setup.git` should get assimilated. The code has been [merged](https://github.com/civicrm/civicrm-core/pull/16618) into `civicrm-core.git`, so the docs should go to `civicrm-dev-docs.git`.

~~This PR is currently WIP - some of the content should be rephrased/reconsidered for the new position.~~

### DIFF
--- a/docs/framework/setup/getting-started.md
+++ b/docs/framework/setup/getting-started.md
@@ -1,0 +1,195 @@
+# Getting started
+
+Whether your goal is to improve the `civicrm-setup` library, write a new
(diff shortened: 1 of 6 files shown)


totten:
@MikeyMJCO @seamuslee001 - I've done a bunch of copy-edits for the imported docs and it should be in better shape.

Removing WIP label.

PR #773, opened 2020-03-02
Consolidate buildkit requirements (basics/requirements.md, tools/buildkit.md)

totten:
1. I believe the list of system requirements in `requirements.md` and `buildkit.md` were originally copied from the same place, but they've drifted apart arbitrarily.  This reconciles them by:
    
    * Copying some of the details in `buildkit.md` over to `requirements.md`
    * Removing the list from `buildkit.md`
    * Replacing with a hyperlink and a blurb about the importance of system requirements
    
2. When reading the "Generic" section, my eyes glazed over at the list of requirements.  There are really just two instructions (`git clone...` and `./bin/civi-download-tools`). Considering that `civi-download-tools` has some diagnostics built-in (and is patch-welcome for more), it seems to belabor the point by reproducing the entire list under `buildkit.md#other-platforms`.

### DIFF
--- a/docs/basics/requirements.md
+++ b/docs/basics/requirements.md
@@ -17,9 +17,9 @@ There are many ways to install MySQL, PHP, and other dependencies -- for example
 
 Civi development should work with most packages -- with a priviso: ***the command-line must support standard command names*** (eg `git`, `php`, `node`, `mysql`, `mysqldump`, etc).
 
-Some environments (e.g. most Linux distributions) are configured properly out-of-the-box.
+Some environments (e.g. most Linux distributions) are configured properly out-of-the-box. Other environments (e.g. MAMP and XAMPP) may require configuring the `PATH`.
 
-Other environments (e.g. MAMP and XAMPP) may require configuring the `PATH`. (See, e.g., [Setup Command-Line PHP](/standards/php.md).)
+<!-- FIXME: There should be a link about diagnosing/fixing paths for third-party binaries. TLDR: `find / -name php -executable` and then update `PATH` via bashrc/bash_profile/whatever -->
 
 ## Buildkit
 
--- a/docs/basics/requirements.md
+++ b/docs/basics/requirements.md
@@ -2,23 +2,24 @@
 
 ## Languages and Services
 
--   Unix-like environment (Linux, OS X, or a virtual machine)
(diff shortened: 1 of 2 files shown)


seamuslee001:
Looks good to me

MikeyMJCO:
Same - happy for a merge on this when you're ready Tim!

totten:
Oh, I'm happy to merge now. :) 

There is a FIXME, and I had taken a stab at that, but I basically got pulled into this sort of icky area:

* You need to know how to maintain your `PATH` or `.bashrc` or `.bash_profile` or whatever for multiple reasons: (1)  for setting up PHP/MySQL in MAMP/XAMPP/etc for CLI usage (which is the FIXME item) and (2) for setting up `~/buildkit` and (3) for the READMEs for various CLI tools (`cv`, `civix`, `git-scan`, etc)
* Different OS vendors and different practitioners on the StackExchange/interwebz point people to doing this in trivially-different ways (`$HOME/.bashrc` vs` $HOME/.profile` vs $HOME/.bash_profile` vs `/etc/profile` vs `/etc/profile.d` vs `/etc/bash.bashrc.d` ad nauseum). This sucks for documenters...  if you write simple instructions ("Put line X in file Y"), then it's guaranteed that some portion of the audience gets confused because their other vendor/friend/tutorial said to do it another way.
* Dealing with bash config should be a standard skill for anyone who uses bash. All of our docs which mention `PATH` should just say "Update your `PATH`" - and let the user decide what exactly that means.
* But anecdotally, I know many newcomers struggle with that.
* So there should be a document we can hyperlink -- a sort of intermediate-level bash doc. So far I haven't found one that I like. There should be one. We shouldn't have to write this. Maybe I'm too picky.
* I do have enough ego that I could try writing such a doc. I started a draft on my laptop. But I can't figure out where this content could fit in the taxonomy of our dev docs. So then I keep second-guessing if I should change my framing...

Anyway, it got me frustrated, and I just want to wash my hands of that FIXME for now.

I think the changes so far are totally fine to merge despite the FIXME. :)

MikeyMJCO:
We have distro specific install steps - adding path management for Ubuntu/Debian should be trivial - I'll look at PRing that change - we've got this done for Ubuntu 16.04+ already!

PR #782, opened 2020-03-20
Add documentation for WP REST API wrapper

kcristiano:
required by https://github.com/civicrm/civicrm-wordpress/pull/160

### DIFF
--- a/docs/api/interfaces.md
+++ b/docs/api/interfaces.md
@@ -1,6 +1,6 @@
 # API Interfaces
 
-The API has three main interfaces along with the PHP Code that can be used to access the API. 
+The API has three main interfaces along with the PHP Code that can be used to access the API.
 
 ## Javascript {#javascript}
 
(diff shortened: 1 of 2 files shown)


MikeyMJCO:
This needs to go under the v3 folder I would guess?

MikeyMJCO:
Then if/when a v4 implementation is released that should be documented separately as v4/wp-rest.md

kcristiano:
@MikeyMJCO I was conflicted on placement.  Technically yes it might belong under v3, but it's CMS specific so I created a new folder.   I am OK with moving and defer to you for placement.

MikeyMJCO:
I'd say put it in as `api/v3/wp-rest.md` - I'm okay with CMS specific stuff in there :-) I think we should link to it then as well from `api/vp/interfaces.md`

kcristiano:
@MikeyMJCO I moved the file to `/docs/api/v3/wp-rest.md` and took a crack at `interfaces.md`   Looks like I have some white space cleanup that snuck in .  let me know if you wnat me to revert that commit.

MikeyMJCO:
I'm good with this - thanks for the free whitespace cleanup :-D

PR #788, opened 2020-03-31
Expand section on how to choose the correct branch for a PR

demeritcowboy:
It's confusing especially to beginners or someone who doesn't do it very often. And for those same people it's very time consuming to then fight with git to try to "fix it". The current docs are helpful but this adds more explanation to aid getting it right the first time.


eileenmcnaughton:
looks good to me

totten:
Ditto. Merging.

PR #790, opened 2020-04-08
Update hook reference to make it clear that cidRefs is deprecated

eileenmcnaughton:
We deprecated altering the data from the hook when called with cidRefs over a year ago but docs were not updated

Screen Shot 2020-04-08 at 12 55 45 PM


MikeyMJCO:
Seems good to me - I'll merge if you're happy for me to Eileen?

eileenmcnaughton:
Yep @MikeyMJCO  - please do

PR #793, opened 2020-04-13
(Repost) Clarify how to use anchor links with ts

demeritcowboy:
Repost of PR #751 which got borked by #755

### DIFF
--- a/docs/translation/index.md
+++ b/docs/translation/index.md
@@ -99,7 +99,21 @@ The general rules for avoiding errors may be summed up like this:
     ```smarty
     <p>{ts}Hello, world!{/ts}</p>
     ```
-    
+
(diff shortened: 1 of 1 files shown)


demeritcowboy:
Thanks! Just as an FYI which might be related to the upgrades, I received the merged notification email three times. Or maybe it's because it was closed and reopened but it's sending a merged notice when somebody closes and another merged notice when reopened?

MikeyMJCO:
Thanks - I'll keep an eye on it - I only got it once - but I'll log the mails for a few weeks and keep an eye on it!

demeritcowboy:
Ok. And just noticing it doesn't seem to be live? Does it only batch process daily or so?

MikeyMJCO:
It should be processing them immediately - it appears to be getting caught up on clearing a cache folder. Looking into it!

PR #795, opened 2020-04-14
Update hook_civicrm_apiWrappers

artfulrobot:
Update this to tell people they need to implement api3 and api4 versions of wrapper classes.

Relevant chat links
- https://chat.civicrm.org/civicrm/pl/g7za3irhk7rc5jhek149udityo
- https://chat.civicrm.org/civicrm/pl/iwxgz7bbsir4tb6xsm9zc6rezo

### DIFF
--- a/docs/hooks/hook_civicrm_apiWrappers.md
+++ b/docs/hooks/hook_civicrm_apiWrappers.md
@@ -17,8 +17,8 @@ Introduced in CiviCRM 4.4.0.
 
 ## Parameters
 
-- API_Wrapper[] $wrappers - an array of objects which implement the [`API_Wrapper`](#wrapper-class) interface.
-- array $apiRequest - contains keys 'entity', 'action', and params; see [APIv3 Usage](../api/v3/usage.md).
+- `API_Wrapper[]` $wrappers - an array of objects which implement the [`API_Wrapper`](#wrapper-class) interface.
+- `array|Civi\Api4\Generic\AbstractAction` `$apiRequest` - contains keys 'entity', 'action', and params; see [APIv3 Usage](../api/v3/usage.md).
 
 ## Returns
 
@@ -29,7 +29,7 @@ Introduced in CiviCRM 4.4.0.
 The `API_Wrapper` interface specifies two methods:
 
  * `fromApiInput($apiRequest)`  - Allows for modifcation of API parameters before the request is executed.
-   Should return a (possibly modified) $apiRequest array.
+   Should return $apiRequest (possibly modified).
 
  * `toApiOutput($apiRequest, $result)` - Allows for modification of the result before it is returned.
    Should return a (possibly modified) $result array.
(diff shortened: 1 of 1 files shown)


docs/hooks/hook_civicrm_apiWrappers.md
```diff
@@ -29,10 +29,10 @@ Introduced in CiviCRM 4.4.0.
 The `API_Wrapper` interface specifies two methods:
 
  * `fromApiInput($apiRequest)`  - Allows for modifcation of API parameters before the request is executed.
-   Should return a (possibly modified) $apiRequest array.
+   Should return $apiRequest (possibly modified).
 
  * `toApiOutput($apiRequest, $result)` - Allows for modification of the result before it is returned.
-   Should return a (possibly modified) $result array.
+   Should return $apiRequest (possibly modified).
```
MikeyMJCO, line 35:
```suggestion
   Should return $apiRequest (possibly modified).
```

Is this an intentional change in the return value?

artfulrobot: old: returns array
new: returns `$apiRequest` (whatever that is - it might be an array or an api request object)

MikeyMJCO: So both `fromApiInput` and `toApiInput` return an `$apiRequest` array/object?

artfulrobot: oops! fixed

docs/hooks/hook_civicrm_apiWrappers.md
```diff
@@ -29,7 +29,7 @@ Introduced in CiviCRM 4.4.0.
 The `API_Wrapper` interface specifies two methods:
 
  * `fromApiInput($apiRequest)`  - Allows for modifcation of API parameters before the request is executed.
-   Should return a (possibly modified) $apiRequest array.
+   Should return $apiRequest (possibly modified).
 
  * `toApiOutput($apiRequest, $result)` - Allows for modification of the result before it is returned.
    Should return a (possibly modified) $result array.
```
artfulrobot, line 35: I've not actually tested in api4 - *is* it an array?

colemanw:
2 versions of a wrapper aren't always necessary. In v3 the `$apiRequest` is an array and in v4 it's an object implementing `arrayAccess` and in both cases you can access `$apiRequest['params']`. However, it should be noted that the arrayAccess behavior is mostly for backward compatibility and treating it like an object like you are doing in your example is probably best.

While we're updating these docs it should be noted that the hook itself is considered "legacy", according to
https://github.com/civicrm/civicrm-core/blob/5.20/Civi/API/Subscriber/WrapperAdapter.php#L33-L38

```php
/**
 * This is a wrapper for the legacy "API Wrapper" interface which allows
 * wrappers to run through the new kernel. It translates from dispatcher events
 * ('civi.api.prepare', 'civi.api.respond') to wrapper calls ('fromApiInput', 'toApiOutput').
 */
```

MikeyMJCO:
> 2 versions of a wrapper aren't always necessary, but it's a good approach. In v3 the `$apiRequest` is an array and in v4 it's an object implementing `arrayAccess` and in both cases you can access `$apiRequest['params']`. However, it should be noted that the arrayAccess behavior is mostly for backward compatibility and treating it like an object like you are doing in your example is probably best.
> 
> While we're updating these docs it should be noted that the hook itself is considered "legacy", according to
> https://github.com/civicrm/civicrm-core/blob/5.20/Civi/API/Subscriber/WrapperAdapter.php#L33-L38
> 
> ```
> /**
>  * This is a wrapper for the legacy "API Wrapper" interface which allows
>  * wrappers to run through the new kernel. It translates from dispatcher events
>  * ('civi.api.prepare', 'civi.api.respond') to wrapper calls ('fromApiInput', 'toApiOutput').
>  */
> ```

Added a note to the top of the page that it's a "legacy" hook.

colemanw:
That note is a good start but should also explain the new/preferred method to use.

MikeyMJCO:
> That note is a good start but should also explain the new/preferred method to use.

You able to take a crack at expanding it? I have no idea what the replacement method is ;-D

colemanw:
Looks good :)

Issue #800, opened 2020-04-30
Accounts Payable should be Accounts Receivable


artfulrobot:
[This line](https://github.com/civicrm/civicrm-dev-docs/blame/master/docs/financial/financialentities.md#L92) (apols for the "blame" - seems to be the only way I can link to a particular line in a Markdown file) says that when an order is created the total $300 will result in:

>financial trxn record transferring the cost ($300) to accounts payable

I think this should be accounts *receivable* - since it's an amount that we are due to receive, not an amount that we are due to have to pay. Two lines down from that confirms my (rather poor) understanding of bookkeeping because it refers to "the accounts *receivable* transaction" (emphasis mine).

@eileenmcnaughton @JoeMurray

eileenmcnaughton:
@artfulrobot I think you are right - wanna do a PR?

artfulrobot:
sure, will do, just wanted to check I was right first!

artfulrobot:
PR: https://github.com/civicrm/civicrm-dev-docs/pull/801

artfulrobot:
Merged, thanks all.

PR #801, opened 2020-05-04
issue/800: correct accounts payable to accounts receivable

artfulrobot:
Correct simple typing mistake where we had "accounts payable" instead of "accounts receivable".

Please see: https://github.com/civicrm/civicrm-dev-docs/issues/800


eileenmcnaughton:
@MikeyMJCO looks ok

PR #804, opened 2020-05-10
Fix link to documentation on symfony hooks in apiwrappers hook docume…

seamuslee001:
…nation

ping @MikeyMJCO @johntwyman


MikeyMJCO:
Good catch. thought I'd caught them all!

PR #806, opened 2020-05-14
Hooks in Symfony - Discuss the $priority option

totten:
When this was first written, I figured that `$priority` is explained in
upstream docs.  However, the truth is that they don't give any guidance on
*what numbers to use* or *how/whether to spread out the numbers*.

This describes some current examples and suggests a general formulation for
folks in doubt.


mattwire:
This looks good

MikeyMJCO:
Merging - content is good and Matt's give it the seal of approval!

PR #810, opened 2020-05-27
Update Documentation following update to Symfony version in CiviCRM Core

seamuslee001:


docs/hooks/hook_civicrm_container.md
```diff
@@ -10,6 +10,9 @@ extensions, etc.
 !!! tip
     The container configuration will be compiled/cached. The default cache behavior is aggressive. When you first implement the hook, be sure to flush the cache. Additionally, you should relax caching during development. In `civicrm.settings.php`, set `define('CIVICRM_CONTAINER_CACHE', 'auto')`.
 
+!!! note
+    As of CiviCRM version 5.27 CiviCRM now uses Symfony v3.4 or v4.4. There is an important change which is that in v3.3 Symfony services are now considered by default to be private. To set be backward compatiable you just need to add onto after your set definiton set `->setPublic(TRUE)` in your code. This will make the service public as was the default originally in the Symfony 2.x series.
+
```
MikeyMJCO, line 15:
```suggestion
!!! note
    As of CiviCRM version 5.27 CiviCRM now uses Symfony v3.4 or v4.4. There is an important change which is that in v3.3 Symfony services are now considered by default to be private. To ensure backwards compatibility you just need to add `->setPublic(TRUE)` after your definition in your code. This will make the service public as was the default originally in the Symfony 2.x series. If you don't need the service to be public, you can omit this.

```

MikeyMJCO:
May need to bump the milestone to 5.28. Either way I'm okay with merging this when we cut the RC for 5.27 or .28.

PR #811, opened 2020-05-29
Update civix.md

PratikDevlekar:
Removed $ from bash commond, "$" makes problem while copying the command.


MikeyMJCO:
So I think the logic here is that the `$` stops people copying commands "blindly" but I agree with you that it's not super-helpful and this is a guide aimed at developers.

Happy to merge and I'll also update the style guide.

PR #812, opened 2020-05-31
Document how to implement cancel recur for payment processors

mattwire:
Documentation for https://github.com/civicrm/civicrm-core/pull/17430


mattwire:
@MikeyMJCO I'm merging this because there is important information from 5.25. We can improve/modify later

PR #814, opened 2020-06-08
Update hook_civicrm_alterLogTables - Fix unclear statement regarding …

stesi561:
…log table conversion

This seemed to be missing a word (INNODB)  - but I've reworked slightly to hopefully avoid confusion between Archive tables (or tables with Archive engine) and Civicrm log tables.


stesi561:
On further consideration - I'm actually wondering if this might be better removed? Doesn't this require @eileenmcnaughton extension to be installed for this to happen.

PR #817, opened 2020-06-11
Remove incorrect statement about empty

eileenmcnaughton:
The documentation implies that empty() will implicitly break. This is not true - empty works on the property bag. However, there is a gotcha - if we look at the below:

Screen Shot 2020-06-12 at 8 24 37 AM

We see that the parameter 'first_name' and 'billing_first_name' have been passed in. So ```$params['firstName'])``` is FALSE for the $params but TRUE for the propertyBag - because the mapping of 'firstName' is defined for the propertyBag.

Conversely - the property bag does not have a mapping for 'billing_first_name' - so it shows as empty. We need to find the params that aren't & should be mapped, like billing_first_name but in terms of docs I don't think we should imply empty 'doesn't work'

(note $g is false above - hard to tell in that screenshot)


MikeyMJCO:
I have no idea about this :-D @mattwire happy that you merge this based on your review!

eileenmcnaughton:
@mattwire so let's flesh it would to where it doesn't work - since it DID work in my testing

mattwire:
@eileenmcnaughton I actually found this hard to reproduce again and started to doubt myself. But I have now reproduced:
image

You can see `$test5` is incorrectly returning TRUE when $params['testvar'] is actually set to "hi". Which makes `empty()` at best unreliable. It's *possible* we can fix this by updating `offsetGet`/`offsetExists` methods but I don't know - @artfulrobot ?

eileenmcnaughton:
@mattwire @artfulrobot good sleuthing! I think that fix by @artfulrobot ^^ will save much pain.

eileenmcnaughton:
Resolved through the other PR

mattwire:
We still need to add a note to the docs that empty will give the wrong results on custom properties in releases earlier than 5.28.

PR #818, opened 2020-06-12
Remove suggestion that you *must* use the new parameters

eileenmcnaughton:
The docs were incorrectly stating that *only* the new properties are available - as can be demonstrated in this tested code in core - https://github.com/civicrm/civicrm-core/pull/17585/files#diff-925c5486a3ef170b6613caf3da110a18R627 - the legacy params still work and while we recommend change no action is required

@mattwire @artfulrobot in my tests the propertyBag is doing what it should here so we should not say that only the new params will work.

Note this will clash with my other docs PR regarding empty not working - if one is merged I will fix the other but I think both discussions need to be open as the docs are currently saying things I can't replicate.


docs/extensions/payment-processors/create.md
```diff
@@ -106,7 +106,7 @@ Best practice right now would be to:
...
 
 !!! note
-    From CiviCRM 5.25 `cancelSubscription()` will receive a `\Civi\Payment\PropertyBag` and *not* an array. This means that if you use any `array_` functions or `empty()` to check parameters your implementation will stop working. Also you will *only* receive a standard set of parameters as defined in `\Civi\Payment\PropertyBag` and will need to update to check for those instead of any you were using.
+    From CiviCRM 5.25 `cancelSubscription()` will receive a `\Civi\Payment\PropertyBag` and *not* an array. This means that if you use any `array_` functions or `empty()` to check parameters your implementation will stop working. Also you will receive a standard set of parameters as defined in `\Civi\Payment\PropertyBag`. The PropertyBag does provide mapping of these standard parameters to legacy parameters but we recommend you switch your usage to the new method. For example the recommended way to access the field civicrm_contribution_recur.processor_id from the propertyBag is $bag->getRecurProcessorID(). However, as of writing Authorize.net is using $params['subscriptionId'] to access the same value and the propertyBag is returning the same value.
```
artfulrobot, line 109: This is fine, but I'd suggest not referring to Authorize.net's code specifically. It might be more useful for devs to link to [the list of legacy params (as of 5.26)](https://lab.civicrm.org/dev/core/-/blob/5.26/Civi/Payment/PropertyBag.php#L32) so they can see for themselves which properties have legacy aliases.

mattwire: Yes. It is actually only once converting back to an array that you are restricted to the new parameters. However we should not be included a reference to Authorize.net and any example should be specific - the reader has no idea who or when it was written and it means nothing to them.
A link to the list of legacy params (as of a specific version of CiviCRM) is a good idea. And perhaps an example. `processor_id` is possibly not the best but also not a bad example because it had a completely unrelated name (ie. `subscriptionId`).
Something like (in a separate paragraph so the information is clearly separated):

For example if your payment processor currently looks for `$params['subscriptionId']` it will continue to work but should be replaced with `$params->getRecurProcessorID()` and the new standardized property name is `recurProcessorID`.
To check if it has a value you would use `if ($params->has('recurProcessorID')) { .. }`

eileenmcnaughton: OK - so remove 

> This means that if you use any `array_` functions or `empty()` to check parameters your implementation will stop working. Also you will *only* receive a standard set of parameters as defined in `\Civi\Payment\PropertyBag` and will need to update to check for those instead of any you were using."

And say

>For most processors this will require no change. However, if you are using array_ functions to interact with the $params/PropetyBag they will no longer work as it is not an array. In addition we recommend you switch your code to take advantage of the PropertyBag. For example 

```
$processorID = $params['subscriptionId'];
```
Would become 
```
$processorID = $params->getRecurProcessorID();
```

To check if it has a value you would use if ($params->has('recurProcessorID')) { .. }

artfulrobot: :+1: looks good to me.

eileenmcnaughton: I updated the PR

mattwire:
Looks good - thanks @eileenmcnaughton @artfulrobot

PR #827, opened 2020-06-29
Update list.md

lcdservices:
Add new customPre hook to listing.


MikeyMJCO:
This is handled by a script and needs to be part of the same PR otherwise we'll end up with build errors.

PR #834, opened 2020-07-21
Docs for CiviReport outputhandlers

demeritcowboy:
WIP docs for https://github.com/civicrm/civicrm-core/pull/17901


MikeyMJCO:
Tag me for review when you're ready.

demeritcowboy:
I think it's ready now.

I wasn't sure about the last line, or even if including it whether that lab space is where it should live.

As an aside I was able to get mkdocs installed for previewing, but the only way I could get php code hiliting to work was with the 3 backticks and specifying php, otherwise no matter I did it seemed to be random hiliting or for a language I didn't recognize, possibly turtle-logo 🐢 .

seamuslee001:
Looks good from my end @MikeyMJCO

docs/hooks/hook_civicrm_alterReportVar.md
```diff
@@ -10,12 +10,23 @@ This hook is used to add or modify display columns and filters.
...
-    "sql", depending on where the hook is called.
+-   $varType - a string containing the value "columns", "rows", "sql",
+    "actions", or "outputhandlers" depending on where the hook is called.
 -   &$var - a mixed var containing the columns, rows, or SQL, depending
     on where the hook is called
 -   $reportForm - a reference to the CRM_Report_Form object.
 
+### Explanation of varType values
+
+-   columns - this is called early in report processing and allows you to add/change the columns in the report. `$var` will contain the existing `$reportForm->_columns` member.
```
MikeyMJCO, line 21: Bullets should use `*` followed by a single space please :-)

docs/hooks/hook_civicrm_alterReportVar.md
```diff
@@ -98,3 +109,111 @@ columns appropriately.
         }
       }
     }
+
+## Example 2
+
+Implementing $varType 'actions' and 'outputhandlers'.
+
+``` php
+    <?php
```
MikeyMJCO, line 118: If using fenced code blocks ` ``` language ` don't also indent the whole code block.

docs/hooks/hook_civicrm_alterReportVar.md
```diff
@@ -98,3 +109,111 @@ columns appropriately.
...
+        // report classname, or the instance id which you can get from
+        // $form->getID(), but it should also be related somehow to what
+        // you used in the $varType == 'actions' hook.
+        return ($form->getOutputMode() === 'sample');
+      }
+
+      // override other functions based on what this format is ...
+    }
+
+    // There is also a working sample extension at
```
MikeyMJCO, line 217: The sample link should be outside the code block and a [text-based link](https://example.org) rather than a https://naked.link.example.org :-)

MikeyMJCO:
A few brief comments from the linting pass :-) looks good otherwise!

demeritcowboy:
Thanks updated as per feedback. So this is awaiting release 5.29.

mlutfy:
@MikeyMJCO Just curious: why are you not merging?

demeritcowboy:
It's an interesting point that technically the feature is available to developers now if they're developing against master. I was mistakenly thinking along the lines of end-user features which are more tied to release numbers. So maybe a more general question is should there be a standard in the dev docs for identifying what versions a feature is available or discontinued. Maybe there is one and I'm not aware of it, or I just need to add some text in the docs somewhere "Added in 5.29"?

mattwire:
@demeritcowboy I'd really like to see somewhere a developer orientated list of added/removed/modified per version so you can easily track what version supports what. But it's probably a lot of work..

MikeyMJCO:
> @demeritcowboy I'd really like to see somewhere a developer orientated list of added/removed/modified per version so you can easily track what version supports what. But it's probably a lot of work..

It will be supported soon via meta-data embedded into the docs. The truth really is that I forget which book I'm looking at PRs for and assign to milestones incorrectly :-p

Merging :-)

PR #850, opened 2020-08-30
dev/core#1328 - Clarify non-DAO output encoding strategy

pfigel:
CiviCRM's output encoding strategy offers quite a few footguns. This attempts to make it easier to understand which output encoding strategy is necessary in which scenario when working with Smarty templates.

### DIFF
--- a/docs/security/outputs.md
+++ b/docs/security/outputs.md
@@ -6,29 +6,57 @@ Untrusted data placed in HTML must be [encoded](index.md#encoding) for HTML outp
 
 ### Between tags {:#db-between-tags}
 
+Whether data needs to be output encoded depends primarily on the data source:
+
 #### Database data between tags {:#db-between-tags}
 
-Data which comes out of MySQL has already been [partially encoded for HTML output](inputs.md#input-encoding). This means that when you place this data between HTML tags, you don't need to perform any output encoding. For example:
+Data which comes *directly* out of MySQL has already been [partially encoded for HTML output](inputs.md#input-encoding). This means that when you place this data between HTML tags, you don't need to perform any output encoding. For example:
+
+```php
+// in page class
+$dao = new CRM_Core_DAO();
+$dao->query('SELECT display_name FROM civicrm_contact');
+if ($dao->fetch()) {
+  // set template variable
+  $this->assign('displayName', $displayName);
+}
+```
 
 ```html
+<!-- in template -->
 <div>{$displayName}</div>
 ```
 
-#### Direct user input between tags {:#inputs-between-tags}
+#### API data or direct user input between tags {:#inputs-between-tags}
 
-Here we have a bit of a grey area where CiviCRM does not have a consistent approach. If untrusted inputs are placed into HTML before being saved to the database, you need to ensure to perform HTML output encoding *at some point*.
-
-You can perform the output encoding in PHP as follows:
+HTML output encoding needs to be performed *at some point* for any data that is fetched via the API as well as any untrusted user input that is placed into HTML before being saved to the database. The recommended approach is to perform output encoding in the template:
 
 ```php
-$userInput = htmlentities($userInput);
+// in page class
+$contacts = \Civi\Api4\Contact::get()
+  ->addSelect('display_name')
+  ->execute();
+foreach ($contacts as $contact) {
+  $this->assign('displayName', $contact['display_name']);
+}
 ```
 
-Or you can perform the output encoding in Smarty as follows:
-
 ```html
-<div>{$userInput|escape}</div>
-``` 
+<!-- in template -->
+<div>{$displayName|escape}</div>
+```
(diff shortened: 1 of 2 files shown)


pfigel:
ping @seamuslee001 @eileenmcnaughton @totten

docs/security/outputs.md
```diff
@@ -6,29 +6,57 @@ Untrusted data placed in HTML must be [encoded](index.md#encoding) for HTML outp
...
 <div>{$displayName}</div>
 ```
 
-#### Direct user input between tags {:#inputs-between-tags}
+#### API data or direct user input between tags {:#inputs-between-tags}
 
-Here we have a bit of a grey area where CiviCRM does not have a consistent approach. If untrusted inputs are placed into HTML before being saved to the database, you need to ensure to perform HTML output encoding *at some point*.
-
-You can perform the output encoding in PHP as follows:
+HTML output encoding needs to be performed *at some point* for any data that is fetched via the API as well as any untrusted user input that is placed into HTML before being saved to the database. The recommended approach is to perform output encoding in the template:
```
pfigel, line 32: I've changed the language a bit here to explicitly recommend encoding in the template rather than in page/form code, as that seems a bit cleaner and more like what you'd typically see in modern frameworks.

MikeyMJCO:
@seamuslee001 could you review?

docs/security/outputs.md
```diff
@@ -6,29 +6,57 @@ Untrusted data placed in HTML must be [encoded](index.md#encoding) for HTML outp
 
 ### Between tags {:#db-between-tags}
 
+Whether data needs to be output encoded depends primarily on the data source:
+
 #### Database data between tags {:#db-between-tags}
 
-Data which comes out of MySQL has already been [partially encoded for HTML output](inputs.md#input-encoding). This means that when you place this data between HTML tags, you don't need to perform any output encoding. For example:
+Data which comes *directly* out of MySQL has already been [partially encoded for HTML output](inputs.md#input-encoding). This means that when you place this data between HTML tags, you don't need to perform any output encoding. For example:
```
seamuslee001, line 13: @pfigel if i run `INSERT INTO civicrm_contact ('display_name') VALUES ('<p onmouseover=alert(\'hi\')>I am evil</p>');`  Then it isn't encoded in the database right? So its not just about the data source but how the data got in there in the first place right?

docs/security/outputs.md
```diff
@@ -6,29 +6,57 @@ Untrusted data placed in HTML must be [encoded](index.md#encoding) for HTML outp
...
+}
 ```
 
-Or you can perform the output encoding in Smarty as follows:
-
 ```html
-<div>{$userInput|escape}</div>
-``` 
+<!-- in template -->
+<div>{$displayName|escape}</div>
```
seamuslee001, line 46: @pfigel I wonder if we would be advocating `|purify` here rather than Escape? Maybe make it as a suggestion instead it probably depends on the type of field right, a rich text field we probably want purify where was a simple text field is probably more about `|escape` right?

docs/security/outputs.md
seamuslee001: Agreed, Do you think it is worth while pointing out that some frameworks I think AngularJS and maybe even current Angular / React do auto sanitization but others like Backbone need explicit escaping / sanitization?

docs/security/outputs.md
```diff
@@ -6,29 +6,57 @@ Untrusted data placed in HTML must be [encoded](index.md#encoding) for HTML outp
...
+```
+
```
seamuslee001, line 48: Also do we need examples in other frameworks?

docs/security/outputs.md
pfigel: This is an interesting point. Civi has this contract where [inputs are encoded](https://docs.civicrm.org/dev/en/latest/security/inputs/#input-encoding) prior to storage, so arguably, the vulnerability here would be in the code bypassing the input coder (even if that's a silly approach). It is something that's worth pointing out - I'll try to work in some language to make clear that only data that has previously gone through the input coder is "secure" in this context.

docs/security/outputs.md
pfigel: My gut feeling is that we should suggest `|escape` by default (matching the default behaviour of modern frameworks), and perhaps mention `|purify` as an alternative for rich text.

docs/security/outputs.md
pfigel: There's the AngularJS section below which mentions how you *shouldn't* use it (but doesn't give an example of how you *should* use it, which would be good to add). Is that what you were thinking about, or something else?

seamuslee001: Yeh something like that I suppose (just was looking at the diff rather than the full page)

homotechsual:
Merged on GitLab.

PR #852, opened 2020-09-10
moved !tip to suit workflow

MarshCastle:
Tendancy will be to run script before reading tips - moving them up puts the warning before the action.


MikeyMJCO:
Good change - thanks!

MarshCastle:
Thanks Mikey :D.

